PR #9, opened 2019-02-20
Add highlighting of active menu items

knuton:
Need to use Mustache in non-strict mode, otherwise all flags for menu entries need to be passed at any time to avoid runtime errors. This has the downside that now also lookups on missing variables fail silently.

### DIFF
--- a/controller/gui/style.css
+++ b/controller/gui/style.css
@@ -23,8 +23,11 @@ dl > dt {
 }
 
 .menu-item-icon:hover {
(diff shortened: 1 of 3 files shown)


pukkamustard:
In my opinion not worth the downside, especially considering that now all pages use a `<h1>` tag indicating page location.

pukkamustard:
Had it running for a bit and then switched back to without the active highlighting. It does make a difference.

PR #18, opened 2019-09-12
Enable connman online check for service selection

knuton:
Enabling connman's online check much improves connman's ability to select the service to use for the default route (Internet access).

This is important for selecting the right service within the same technology (double Ethernet), where otherwise no natural hierarchy exists.

Compare the outcome of the following scenario with online check off vs on:

**Scenario:** PlayOS starts while only connected to the Senso, other Ethernet interface is unplugged or otherwise physically down. After startup, the Senso-bound interface is assigned a link-local address and assumes READY state in connman.

**Connecting Internet-bound interface WITH online check:** The second Ethernet interface comes up and is tested for online access. The test succeeds and the service assumes ONLINE state, causing it to be prioritized over the existing service in READY state.

**Connecting Internet-bound interface WITHOUT online check:** The second Ethernet interface comes up and assumes READY state. There is no reason to prioritize it over the existing service, which is also READY, and the default route is not changed.

We initially disabled the online check, seeing no concrete benefit and preferring to do an application-level online check instead. Clearly, there IS a concrete payoff to performing the check in the network
management level, as seen from the above.

Marcel Holtmann of Intel makes the following, sensible, argument as to why the online check is reasonably designed for privacy:

    If everybody uses the same service then it is fully anonymous since
    nobody can tell you apart from the other ConnMan on your network. If
    we start fragmenting, then this looses its anonymous status where
    everybody is equal.

https://lists.01.org/pipermail/connman/2015-July/020255.html
https://lists.01.org/pipermail/connman/2015-July/020254.html

### DIFF
--- a/system/networking/default.nix
+++ b/system/networking/default.nix
@@ -23,8 +23,8 @@
         # Allow simultaneous connection to ethernet and wifi
         SingleConnectedTechnology=false
 
-        # Disable calling home
-        EnableOnlineCheck=false
+        # Enable online check to favour connected services
+        EnableOnlineCheck=true
       '';
     };
     # Issue 1: Add a dummy network to make sure wpa_supplicant.conf


krksgbr:
Minimal information is sent to the server, and the code is open-source, so it's verifiable.
An unlikely future scenario is still technically possible, where a new version of ConnMan would start sending more information, so there is still some trust involved, but the risk seems negligible.

knuton:
> Is the application online check still useful now ? Is there some cleanup to do ?

I think this is useful as is in the kiosk application. ConnMan has built-in and sophisticated [support for captive portal detection](https://git.kernel.org/pub/scm/network/connman/connman.git/tree/src/wispr.c?h=1.37#n825), but in the end it just requests to [launch a browser](https://git.kernel.org/pub/scm/network/connman/connman.git/tree/doc/agent-api.txt?h=1.37#n36). I am not convinced we need its level of sophistication (we don't need actual WISPr), our check should cover the majority of cases, and we don't couple ourselves to a particular network manager in that way.

The only place we might reconsider is the check in the playos-controller. Instead of the check here, we could rely on information from ConnMan. Another idea however, is to only manage connection to the network _as such_ in the controller, and let the kiosk do its job with respect to captive portals. This would mean removing the online check functionality from the controller. I suggest to go forward with the current implementation and revisit this with further input from field use as needed.

> An unlikely future scenario is still technically possible, where a new version of ConnMan would start sending more information, so there is still some trust involved

Even rogue updates in general are possible, but ConnMan will continue to send the same amount of information until we update to a hypothetical spying version. Should we realize that a new version requires more data to be exchanged, we can set up our own endpoint and patch the sources.

knuton:
> I discovered it when I was working on the captive portal implementation, but was not convince by the usage of opening an external browser.

We can [hook it to our own action](https://github.com/dividat/playos/blob/2019.8.0/controller/bindings/connman/connman.ml#L115-L120), but it still requires inter-process communication in a way the current solution does not.

etaque:
Inline with your take on this.

PR #19, opened 2019-09-13
Show PlayOS version in user agent

guyonvarch:
Tested on VM with the following modifications:

```diff
diff --git a/system/play-kiosk.nix b/system/play-kiosk.nix
index 73b178b2..c0f27a45 100644
--- a/system/play-kiosk.nix
+++ b/system/play-kiosk.nix
@@ -43,7 +43,7 @@
             export QTWEBENGINE_REMOTE_DEBUGGING="127.0.0.1:3355"
 
             ${pkgs.playos-kiosk-browser}/bin/kiosk-browser \
-              ${config.playos.kioskUrl} \
+              https://www.whatismybrowser.com/detect/what-is-my-user-agent \
               http://localhost:3333/
 
             waitPID=$!
```

Still to be done:

- [x] Cleanup / refactor the code
- [x] Profile the kiosk browser to verify that there are no memory leaks
- [x] Test on a real machine

### DIFF
--- a/kiosk/bin/kiosk-browser
+++ b/kiosk/bin/kiosk-browser
@@ -21,21 +21,13 @@ parser.add_argument(
     'Keyboard combination to toggle between primary and secondary websites. (Default: "%s")'
     % default_toggle_key)
 
-parser.add_argument('system_name', help='System name displayed in user agent')
-parser.add_argument('system_version', help='System version displayed in user agent')
 parser.add_argument('primary_url', help='Primary URL to load')
 parser.add_argument('secondary_url', help='Secondary URL to load')
 
 args = parser.parse_args()
 
 # try:
-kiosk_browser.start(
-        system_name = args.system_name,
-        system_version = args.system_version,
-        primary_url = args.primary_url,
-        secondary_url = args.secondary_url,
-        toggle_sequence = args.toggle_key
-)
+kiosk_browser.start(args.primary_url, args.secondary_url, args.toggle_key)
 
 # except InvalidUrl as error:
 #   parser.error(error)
--- a/kiosk/bin/kiosk-browser
+++ b/kiosk/bin/kiosk-browser
@@ -21,13 +21,14 @@ parser.add_argument(
     'Keyboard combination to toggle between primary and secondary websites. (Default: "%s")'
     % default_toggle_key)
 
+parser.add_argument('name', help='Name displayed in user agent')
 parser.add_argument('primary_url', help='Primary URL to load')
 parser.add_argument('secondary_url', help='Secondary URL to load')
 
 args = parser.parse_args()
 
 # try:
-kiosk_browser.start(args.primary_url, args.secondary_url, args.toggle_key)
+kiosk_browser.start(args.name, args.primary_url, args.secondary_url, args.toggle_key)
 
 # except InvalidUrl as error:
 #   parser.error(error)
--- a/kiosk/bin/kiosk-browser
+++ b/kiosk/bin/kiosk-browser
@@ -21,14 +21,21 @@ parser.add_argument(
     'Keyboard combination to toggle between primary and secondary websites. (Default: "%s")'
     % default_toggle_key)
 
-parser.add_argument('name', help='Name displayed in user agent')
+parser.add_argument('system_name', help='System name displayed in user agent')
+parser.add_argument('system_version', help='System version displayed in user agent')
 parser.add_argument('primary_url', help='Primary URL to load')
 parser.add_argument('secondary_url', help='Secondary URL to load')
 
 args = parser.parse_args()
 
 # try:
-kiosk_browser.start(args.name, args.primary_url, args.secondary_url, args.toggle_key)
+kiosk_browser.start(
+        system_name = args.system_name,
+        system_version = args.system_version,
+        primary_url = args.primary_url,
+        secondary_url = args.secondary_url,
+        toggle_sequence = args.toggle_key
+)
 
 # except InvalidUrl as error:
 #   parser.error(error)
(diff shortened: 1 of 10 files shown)


kiosk/kiosk_browser/browser_widget.py
```diff
@@ -1,13 +1,19 @@
...
 
 class BrowserWidget(QWebEngineView):
-    def __init__(self, url, *args, **kwargs):
+
+    def __init__(self, name, url, *args, **kwargs):
         QWebEngineView.__init__(self, *args, **kwargs)
 
-        self.load(url)
+        profile = QWebEngineProfile.defaultProfile()
+        profile.setHttpUserAgent(f"{name} {profile.httpUserAgent()}")
```
knuton: I think it would be nice to stick to the [conventional format](https://en.wikipedia.org/wiki/User_agent#Format_for_human-operated_web_browsers) here, which is something like `Mozilla/5.0 (iPad; U; CPU OS 3_2_1 like Mac OS X; en-us) AppleWebKit/531.21.10 (KHTML, like Gecko) Mobile/7B405`. There is also a panoply of user agent strings using this format in the [platform.js test suite](https://github.com/bestiejs/platform.js/blob/master/test/test.js).

This implies replacing the contents of the first set of parentheses (which is "Details of the system in which the browser is running") with something like `(X11; PlayOS 19.8.0)`.

guyonvarch: I chose to kept the Linux part of the system, that’s not bothering and it adds extra information.

`Mozilla/5.0 (X11; Linux x86_64; PlayOS 2.1.0) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.12.0 Chrome/69.0.3497.128 Safari/537.36`

I added a fallback in case there is something wrong with the matching, in this case, the PlayOS version is set at the begining of the user agent.

kiosk/kiosk_browser/browser_widget.py
```diff
@@ -29,9 +32,38 @@ def __init__(self, url, *args, **kwargs):
         self.setSizePolicy(policy)
 
     def clean_and_load(self, url):
-        self.setPage(QWebEnginePage())
-        self.load(url)
+
+        # Recreate a new page in order to clear the screen the moment the
+        # shortcut is pressed
+        page = QWebEnginePage(self._profile)
```
guyonvarch: We’re recreating a page instance each time, instead of reusing the same one. I added a comment to indicate why we do this.

From [Qt documentation](https://doc.qt.io/qt-5/qwebengineview.html#setPage), I understand that `setPage` will free the current page when loading a new page:

>  If the current page is a child of the web view, it will be deleted.

I’ll profile the kiosk browser locally by changing the urls a lot following [this guide](https://medium.com/zendesk-engineering/hunting-for-memory-leaks-in-python-applications-6824d0518774)

kiosk/kiosk_browser/browser_widget.py
guyonvarch: Memory is not freed when recreating a new page instance each time:

memory-profile-new-page-instance

No memory leak when using the same page instance:

memory-profile-same-page-instance

guyonvarch:
The memory is leaking on `develop` because of the page creation at url load:

memory-profile

It was introduced with the captive portal PR.

https://github.com/dividat/playos/blob/cbb8e2413fe0686e1da7c4700878931d7165e79e/kiosk/kiosk_browser/browser_widget.py#L31-L33

Luckily, this change is neither in `master` nor in `validation`.

guyonvarch:
I tested the change on a real machine, logged in to a captive portal. The version is showed in the user agent.

kiosk/bin/kiosk-browser
```diff
@@ -21,13 +21,21 @@ parser.add_argument(
     'Keyboard combination to toggle between primary and secondary websites. (Default: "%s")'
     % default_toggle_key)
 
+parser.add_argument('system_name', help='System name displayed in user agent')
+parser.add_argument('system_version', help='System version displayed in user agent')
```
knuton: What's the motivation of splitting into system name and system version?

It allows to fall back to `<name>/<version> <default-user-agent>` in case parsing fails, but it also complicates the implementation, and even forces specifying something for both name and version.

Could the caller decide which system identifier to use?

kiosk/default.nix
```diff
@@ -11,6 +11,12 @@ python3Packages.buildPythonApplication rec {
     (path: type: type != "directory" ||  baseNameOf path != "venv")
     ./.;
 
+  postPatch = ''
+    substituteInPlace kiosk_browser/system.py \
+      --replace @name@ ${system_name} \
```
knuton: Assumes `system_name` never includes whitespace.

guyonvarch: Well spotted!

I considered using `substituteAllInPlace`:

```diff
diff --git a/kiosk/default.nix b/kiosk/default.nix
index 811797e8..3e0dba30 100644
--- a/kiosk/default.nix
+++ b/kiosk/default.nix
@@ -6,15 +6,14 @@ let qtx=qt5; in
 python3Packages.buildPythonApplication rec {
   pname = "playos_kiosk_browser";
   version = "0.1.0";
+  inherit system_name system_version;
 
   src = builtins.filterSource
     (path: type: type != "directory" ||  baseNameOf path != "venv")
     ./.;
 
   postPatch = ''
-    substituteInPlace kiosk_browser/system.py \
-      --replace @name@ ${system_name} \
-      --replace @version@ ${system_version}
+    substituteAllInPlace kiosk_browser/system.py
   '';
 
   doCheck = false;
diff --git a/kiosk/kiosk_browser/system.py b/kiosk/kiosk_browser/system.py
index 472a5445..f5386ade 100644
--- a/kiosk/kiosk_browser/system.py
+++ b/kiosk/kiosk_browser/system.py
@@ -1,2 +1,2 @@
-NAME = "@name@"
-VERSION = "@version@"
+NAME = "@system_name@"
+VERSION = "@system_version@"
```

But that’s brittle, someone could remove the `inherits` without noticing and we would loose the substitution. So, I sticked to `substituteInPlace`.

PR #20, opened 2019-09-24
Use breeze contrast cursor theme to make the cursor more visible

guyonvarch:
## Inspirations

- AUR package:
https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=breeze-contrast-cursor-theme


knuton:
Looks good.

I suggest moving the theme derivation definition to [`pkgs/`](https://github.com/dividat/playos/tree/develop/pkgs) with the other custom packages, would be slightly cleaner still.

guyonvarch:
> I suggest moving the theme derivation definition to pkgs/ with the other custom packages, would be slightly cleaner still.

Sure, done!

knuton:
Very well.

PR #22, opened 2019-11-08
Clean connection state after a connection failure

guyonvarch:
Previously, when an error happened during a connection, the agent was
not cleaned up. At the next connection attempt, an Already Existed error
was returned. A reboot of PlayOS was required to be able to connect to a
wifi network.

The proposed change clean up the agent, no matters if the connection
succeeds or fail. It allows to reconnect without restarting PlayOS.


knuton:
Good, this comes up occasionally.

PR #23, opened 2019-11-11
Show service strength alongside its name in networks

guyonvarch:
The strength was already accessible in each service detail, but it was
too hidden to get this information at a glance.

When the strength of a network is too low, the user knows he will have
difficulties connecting to it. It can be a trigger for him to take
action to get a better strength if possible.

### DIFF
--- a/controller/bindings/connman/connman.ml
+++ b/controller/bindings/connman/connman.ml
@@ -328,6 +328,9 @@ struct
     ; "name", s.name |> Ezjsonm.string
     ; "favorite", s.favorite |> Ezjsonm.bool
     ; "connected", s |> is_connected |> Ezjsonm.bool
+    ; "strength", (match s.strength with
+      | Some s -> string_of_int s ^ "%"
+      | None -> "") |> Ezjsonm.string
     ; "properties", s |> sexp_of_t |> Sexplib.Sexp.to_string_hum |> Ezjsonm.string
     ]
 
--- a/controller/gui/style.css
+++ b/controller/gui/style.css
@@ -58,10 +58,6 @@ dl > dt {
     margin: 10px;
 }
 
-.service-detail {
-    margin-left: 10px;
-}
-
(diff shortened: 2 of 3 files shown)


controller/bindings/connman/connman.ml
```diff
@@ -328,6 +328,9 @@ struct
     ; "name", s.name |> Ezjsonm.string
     ; "favorite", s.favorite |> Ezjsonm.bool
     ; "connected", s |> is_connected |> Ezjsonm.bool
+    ; "strength", (match s.strength with
+      | Some s -> string_of_int s ^ "%"
```
knuton: Are you sure that this value is in percent? Just wondering because all networks listed seem rather low in strength.

controller/gui/style.css
```diff
@@ -85,3 +81,14 @@ dl > dt {
 
     height: 90vh;
 }
+
+/* Network */
+
+.d-Network__ServiceDetail {
```
knuton: Not sure whether to use BEM and start a mixed style here.

Since the project is rather small, I would either

- stick to the existing primitive style, or
- swap everything to BEM style for consistency.

controller/bindings/connman/connman.ml
guyonvarch: Yes, it’s in percent. Moving to nearer to the Wifi access point gives me this:

guyonvarch:
> An additional color grading scheme in spectrum [red, green] might be nice, but can be misleading if most networks show up as below e.g. 70 % in practice. Maybe a later enhancement.

I thought about it, but was not sure neither. And it could be bars too. I would opt to let it for a later enhancement.

controller/gui/style.css
guyonvarch: Opting to swap everything to BEM style for consistency. It will not be long because there are few styles. I’ll create another issue.

PR #24, opened 2019-11-11
Reload network page with a button

guyonvarch:
This allows the user to get a fresh network list when he clicks on the
reload button.

### DIFF
--- a/controller/gui/style.css
+++ b/controller/gui/style.css
@@ -89,11 +89,11 @@ dl > dt {
 }
 
 .d-Network__ServiceStrength {
-  color: #777777;
-  font-size: 80%;
+    color: #777777;
+    font-size: 80%;
 }
 
-.d-Network__Reload {
+.d-Network__Refresh {
     display: inline-block;
     font-size: 1rem;
     text-decoration: none;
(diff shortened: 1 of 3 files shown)


knuton:
Good to provide a more explicit option for refreshing. I wonder if "Reload" could be confusing to some: What is being reloaded? Could it possibly break something?

Maybe another label would be more reassuring. "Refresh"? "Get Status"? "Refresh List"?

guyonvarch:
> I wonder if "Reload" could be confusing to some: What is being reloaded? Could it possibly break something?
>
> Maybe another label would be more reassuring. "Refresh"? "Get Status"? "Refresh List"?

We could rephase it “Reload the page”, but it may be too verbose. “Refresh List” is partially right, because the connection status is reloaded too. That’s the reason I put the button on top of the informations. “Refresh” fit the intent and is concise, let’s go with it.

controller/gui/style.css
```diff
@@ -85,3 +85,20 @@ dl > dt {
 
     height: 90vh;
 }
+
+/* Network */
+
+.d-Network__Refresh {
```
knuton: For future UI consistency, probably best to give this a generic name so that the same button style could be used in other, similar places in the controller.

guyonvarch: > For future UI consistency, probably best to give this a generic name so that the same button style could be used in other, similar places in the controller.

Right, I’ll make a pass in another PR, there are already some buttons with the default style.

PR #25, opened 2019-11-11
Prevent service does not exist error

guyonvarch:
The proposed solution store all fetched services in a variable.

Previously:

1. When the user asked for the network page, the server fetched services
and returned those inside the page.
2. Then, the user selected one service and tried to connect to it by
sending its id to the server.
3. Finally, the server re-fetched services and tried to find the service
the user wants to connect to by using its id.

The service can not be found in the newly fetched services.

With the proposed modification:

1. When the user ask for the network page, the server fetches services,
store the new ones with the old ones in a mutable list, and return the
new services inside the page.
2. Then, the user select one service and tries to connect to it by
sending its id to the server.
3. Then, the server tries to find the service in the list of stored
services.

Because every all the previous services have been stored, the server
should always find back the service the user wants to connect to.

### Checks

- [ ] Add changelog

### DIFF
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -253,36 +253,45 @@ module NetworkGui = struct
   open Connman
   open Network
 
+  (** Mutable services resulting from the merge of every
+      services ever given from the overview page.*)
+  let merged_overview_services = ref []
+
+  (** Merge new services into old services.*)
+  let merge_services old_services new_services =
+    old_services
+    |> List.filter
+      (fun (o:Service.t) -> List.exists (fun (n:Service.t) -> o.id == n.id) new_services)
+    |> List.append new_services
+
+  (** Find a service by id.*)
+  let find_service id =
+    List.find_opt (fun (s:Service.t) -> s.id = id)
+
   let overview
       ~(connman:Manager.t)
       ~(internet:Internet.state Lwt_react.S.t)
       req =
 
-    (* Check if internet connected *)
+    (* Check if internet is connected *)
     let internet_connected =
         internet
         |> Lwt_react.S.value
         |> Internet.is_connected
     in
 
-    let%lwt services =
-      Manager.get_services connman
-      (* If not connected show all services, otherwise show services that are connected *)
-      >|= List.filter (fun s -> not internet_connected || s |> Service.is_connected)
-    in
+    let%lwt services = Manager.get_services connman in
 
+    (** Store new services.*)
+    let () = merged_overview_services := merge_services !merged_overview_services services in
 
     page "network"
-      [
-        "internet_connected", internet_connected |> Ezjsonm.bool
-      ; "services", services |> Ezjsonm.list (fun s -> s |> Service.to_json |> Ezjsonm.value)
+      [ "internet_connected", internet_connected |> Ezjsonm.bool
+      ; "services", services
+        (* If not connected show all services, otherwise show services that are connected *)
+        |> List.filter (fun s -> not internet_connected || s |> Service.is_connected)
+        |> Ezjsonm.list (fun s -> s |> Service.to_json |> Ezjsonm.value)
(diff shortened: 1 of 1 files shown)


guyonvarch:
The proposed solution does not clean old services from the store. If a wifi point was available and is not available anymore, it will still be present in storage, but will not be available to the user because only fresh and available services are returned in the overview page.

controller/server/gui.ml
```diff
@@ -253,37 +253,46 @@ module NetworkGui = struct
...
 
+  (** Mutable services resulting from the merge of every
+      services ever given from the overview page.*)
+  let merged_overview_services = ref []
+
+  (** Merge new services into old services.*)
+  let merge_services old_services new_services =
+    old_services
+    |> List.filter
+      (fun (o:Service.t) -> List.exists (fun (n:Service.t) -> o.id == n.id) new_services)
```
knuton: Is a `not` missing here or am I confused? I assume the intent is to combine previously known services that are no longer found with all services that have now been found.

guyonvarch: Right, I may have been confused by Ocaml equality testing it.

guyonvarch:
> I am a bit worried about the conceptually infinitely growing list. Maybe pruning by timestamp or
a generation counter would still be nice to guarantee the list maintaining a reasonable size.

That would be safer. :+1:

> Another thing I am wondering is whether we would duplicate some battle-tested functionality implemented in connman/wpa_supplicant. Unfortunately the docs of both projects are not very informative on this issue, and the sources hard to make sense of for me.

Thanks for pointing this, I’ll search if it can give the same result.

guyonvarch:
I tested `wpa_supplicant` options.

This diff reports missing networks that were previously returned:

```diff
diff --git a/controller/server/gui.ml b/controller/server/gui.ml
index 42300fcb..58bdd440 100644
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -261,7 +261,11 @@ module NetworkGui = struct
   let merge_services old_services new_services =
     old_services
     |> List.filter
-      (fun (o:Service.t) -> not (List.exists (fun (n:Service.t) -> o.id = n.id) new_services))
+      (fun (o:Service.t) ->
+        let keep = not (List.exists (fun (n:Service.t) -> o.id = n.id) new_services) in
+        let () = if keep then print_endline ("Keeping " ^ o.name) else () in
+        keep
+      )
     |> List.append new_services

   (** Find a service by id.*)
```

With default values of `bss_expiration`:

```conf
bss_expiration_age=180
bss_expiration_scan_count=2
```

Within less than a minute, there are missing networks reported.

With very high values of `bss_expiration`:

```conf
bss_expiration_age=18000
bss_expiration_scan_count=2000
```

There are no missing networks reported after 15 minutes.

The drawback though is that we can’t set too high values neither, because a dead network would be kept.

We would need to know what is the frequency of scan to determine an expiration scan count that fit our needs.

Reading the `BackgroundScanning` of [connman man page](https://manpages.debian.org/unstable/connman/connman.conf.5.en.html), it may trigger a scan every 3 second when wifi is up, but it’s not explicit.

Setting bss expiration to our needs is still a different resolution that the one proposed on this PR. The one proposed in this PR return fresh services, whereas by increasing the cache on `wpa_supplicant`, less fresh services would be returned.

--

There is a limit of 200 bssid in cache (ap_scan=1 is the default). It seems to be already enough.

```
# Maximum number of BSS entries to keep in memory
# Default: 200
# This can be used to limit memory use on the BSS entries (cached scan
# results). A larger value may be needed in environments that have huge number
# of APs when using ap_scan=1 mode.
#bss_max_count=200
```

guyonvarch:
Closing, https://github.com/dividat/playos/pull/26 is tackling the same problem on `wpa_supplicant` level.

PR #26, opened 2019-11-12
Increase BBS scans expiration

guyonvarch:
This change aims to fix WIFI connection when there are lot of available
WIFI networks.

When loading the network page, the list of WIFI networks are sent to the
user. When he wants to connect to one of them, he sends the id of the
WIFI network he wants to connect to, then, the server fetchs the list of
WIFI networks again and tries to find the user one. But when there are
lot of WIFI networks available, only a subset of available WIFI networks
are returned at each scan. It’s possible that the network the user wants
to connect to is not returned this time.

By default, WIFI networks are discarded if they don’t appear in 2
consecutive scans, see [wpa_supplicant configuration documentation](
https://w1.fi/cgit/hostap/plain/wpa_supplicant/wpa_supplicant.conf). We
could increase the number of required succesive scan before abandoning a
network.

After shutting down a WIFI network, I measured the time it was displayed
in cmst, according to various count of BBS expirations.

- 2 [default]: 25 seconds
- 4: 34 seconds
- 8: 52 seconds
- 16: 1 minute 30 seconds

By requiring 16 consecutive scans before droppinga network, that could
be enough to keep networks in a crowded WIFI place. The drawback is that
if a WIFI network is off, the user will know it 1mn30s after, instead of
25s after.

### Checklist

- [x] Add changelog
- [x] Test on PlayOS

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@
 ## Fixed
 
 - controller: Allow subsequent connections after a connection error
+- controller: Stabilize WIFI connection scanning
 
 # [2019.9.0] - 2019-10-15
 
--- a/system/networking/default.nix
+++ b/system/networking/default.nix
@@ -33,12 +33,12 @@
       extraConfig = ''
         # BSS expiration age in seconds. A BSS will be removed from the local cache
         # if it is not in use and has not been seen for this time. Default is 180.
-        bss_expiration_age=180
+        bss_expiration_age=60
 
         # BSS expiration after number of scans. A BSS will be removed from the local
         # cache if it is not seen in this number of scans.
         # Default is 2.
-        bss_expiration_scan_count=16
+        bss_expiration_scan_count=1000
       '';
 
       # Issue 1: Add a dummy network to make sure wpa_supplicant.conf
--- a/system/networking/default.nix
+++ b/system/networking/default.nix
@@ -27,10 +27,22 @@
         EnableOnlineCheck=true
       '';
     };
-    # Issue 1: Add a dummy network to make sure wpa_supplicant.conf
-    # is created (see https://github.com/NixOS/nixpkgs/issues/23196)
     wireless = {
       enable = true;
+
+      extraConfig = ''
+        # BSS expiration age in seconds. A BSS will be removed from the local cache
+        # if it is not in use and has not been seen for this time. Default is 180.
+        bss_expiration_age=180
+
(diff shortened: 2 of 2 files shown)


knuton:
Nice. I think the trade-off of dangling unavailable wifis would be acceptable. I am thinking that about a minute, so something like 10 consecutive scans might already be enough.

For testing with PlayOS, it might be useful to fire up Wireshark in monitor mode (if available) and look at the beacon frames sent (though I am not sure how active and therefore visible background scanning really is).

guyonvarch:
I tested the modification on a physical PlayOS, after shutting down the WIFI access point, it took 3 minutes for it to disappear. But this duration correspond to the expiration duration. Thinking about it, we are more interested into setting a duration than a scan count, so the solution is to set the expiration duration to 1 minute and set scan expiration to an important value, 1000 for example. I’ll recheck with it.

I activated wireshark on the wifi interface, but I was not able to see scan frames going. I am not used to wireshark, so I may miss something.

When I started PlayOS, I had no WIFI networks showed, though the WIFI interface was listed in system status. I had to restart the computer to have it working. Do you know if this problem already happen on previous versions ? I find it strange that this configuration change would be the cause of it, but who knows.

knuton:
> When I started PlayOS, I had no WIFI networks showed, though the WIFI interface was listed in system status. I had to restart the computer to have it working. Do you know if this problem already happen on previous versions ? I find it strange that this configuration change would be the cause of it, but who knows.

This has been an issue especially on first boot. Related commits with attempted fixes:

- https://github.com/dividat/playos/pull/13/commits/6b929f9d01793bb913871f4ff8549f2b258fb7e1
- https://github.com/dividat/playos/pull/13/commits/2b05d9afc61b7a0e167e96ecd592f2e70dc2035c
- https://github.com/dividat/playos/pull/13/commits/15971f06d5a3d49a7ac430c484e6bd6ac878cce6

The currently applied fix is still based on experimentation rather than deep understanding. But the core issue seems to be a race issue between rfkill blocking the wifi device and wpa_supplicant/connman starting.

guyonvarch:
I tested the expiration config. After shutting down my WIFI network, it disappeared from the list after just a little more than 1 minute.

knuton:
:+1: 

Maybe a brief summary of the problem this configuration addresses would be good to put in a comment.

knuton:
Very well!

Merging this and will test in-depth together with other changes.

PR #27, opened 2019-11-13
Use BEM style in controller

guyonvarch:


Readme.md
```diff
@@ -56,6 +56,10 @@ make develop
 
 When switching key pairs on a channel, the new certficiate must be built into the bundle, which must then be signed with the old key. For this purpose, the `--override-cert` option of the deploy script is needed to provide RAUC with a certificate matching the new key.
 
+## Change Log
+
+Update the change log for every release. See http://keepachangelog.com/ for formatting and conventions.
```
guyonvarch: Would this be a good idea be able to view PlayOS’s change log from the controller ?

knuton: Nice to have, yes.

controller/Readme.md
```diff
@@ -28,7 +28,10 @@ Notes:
 - The controller application requires certain artifacts to be in a specific location relative to binary location. The `dune build @install` command ensures this.
 - The `--profile release` option disables dune from failing on warnings (which are currently present in the `obus` library)
 
+## Code style
+
+- Author CSS according to these [BEM-based guidelines](https://github.com/dividat/diviapps/blob/develop/frontend/docs/dev/styles.md).
```
knuton: I think linking to private resources in a public repo isn't very nice.

Maybe just saying 

```
- Author CSS according to the [BEM methodology](http://getbem.com/) in the format `d-Block__Element--Modifier.
```
will be enough in this context. Enough information for internal devs and if we have external contributors we will have to find another method of knowledge transfer anyway.

controller/gui/style.css
```diff
@@ -1,43 +1,49 @@
...
 
-.hidden-link {
-    text-decoration: none;
-    cursor: text;
-    color: black;
-}
+/* Menu */
 
-#menu {
+#d-Menu {
```
knuton: While at it: I don't see any benefit to using `id` selectors here. These could just be classes.

Readme.md
guyonvarch: I added a card.

controller/Readme.md
guyonvarch: Right!

controller/gui/style.css
guyonvarch: I was not sure if it was BEM related or not, I am also in favor of using classes.

PR #28, opened 2019-11-13
Improve controller UI

guyonvarch:
At first, I just wanted to harmonize forms. But then, I took the opportunity to improve other parts of the controller as well.

- Add reset.css
- Design forms
- Design preformatted text blocks
- Emphasize internet connection
- Add semantic HTML tags

# Screenshots

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,10 @@
 - controller: Add a refresh button in the network page
 - controller: Show network strengths in the network page
 
+## Changed
+
+- controller: Improve UI
+
 ## Fixed
 
 - controller: Allow subsequent connections after a connection error
--- a/controller/dune
+++ b/controller/dune
@@ -2,6 +2,7 @@
  (section share_root)
  (files
   (gui/client/client.bc.js as static/client.js)
+  (gui/reset.css as static/reset.css)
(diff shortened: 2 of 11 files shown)


knuton:
This looks good. Note: It shouldn't look _too_ good or inviting, it is beneficial if it has a slightly technical and low-level look to signal that it is for staff/advanced usage.

I see there is some duplication for the yellow boxes. Can this be avoided? A naïve solution would be to just use a common wrapper element for any content that should have this style.

guyonvarch:
> I see there is some duplication for the yellow boxes. Can this be avoided? A naïve solution would be to just use a common wrapper element for any content that should have this style.

I don’t see where is the duplication. There is the `.d-Preformatted` class:

```css
.d-Preformatted {
    font-size: 80%;
    border-left: 0.5rem solid #E0CB52;
    padding: 0.5rem 1rem;
    background-color: #FCFAEE;
    overflow: auto;
}
```

And it’s used like this every time we need it:

```html
<pre class="d-Preformatted">…</pre>
```

knuton:
Indeed, my bad, sorry. There is duplication of border rules for the various form elements, but that seems tolerable. Though CSS variables might help unify those.

PR #29, opened 2019-11-28
Add wpa_supplicant restart hook on rfkill unblock

knuton:
This uses udev to restart wpa_supplicant whenever a rfkill block of wifi is removed ("unblock"). This seems a bit trigger happy, but is probably the most direct solution as long as the underlying problem in wpa_supplicant/connman's interaction with rfkill is not known. Our machines are also unlikely to toggle rfkill during operation, so the practical downsides should be low.

While the heuristic timer-based restarting of wpa_supplicant mostly works in regular setups, picking a timer duration is difficult: Longer durations are more likely to work (coming after the unblock), but overly long durations are bad for the user and may even lead to mistaken conclusions of there being no connectivity at all.

Point in case, the duration was just about too short for live systems ("lab key"), so they would frequently boot into a state where wifi connectivity could not be restored without a system reboot.

### DIFF
--- a/system/networking/default.nix
+++ b/system/networking/default.nix
@@ -62,6 +62,7 @@
   environment.etc = [
     { source = pkgs.writeScript "rfkill.hook" ''
       #!${pkgs.runtimeShell}
+      # States: 1 - normal, 0 - soft-blocked, 2 - hardware-blocked
       if [ "$RFKILL_STATE" == 1 ]; then
         # Wait an instant. Immediate restart gets wpa_supplicant stuck in the same way.
         sleep 5
--- a/system/networking/default.nix
+++ b/system/networking/default.nix
@@ -55,19 +55,22 @@
   };
   # Issue 2: Make sure connman starts after wpa_supplicant
   systemd.services."connman".after = [ "wpa_supplicant.service" ];
-  # Issue 3: Leave time for rfkill to unblock WLAN and restart wpa_supplicant & connman (https://01.org/jira/browse/CM-670)
-  systemd.timers."restart-wpa" = {
-    timerConfig = {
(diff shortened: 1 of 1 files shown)


system/networking/default.nix
```diff
@@ -55,19 +55,22 @@
...
-    serviceConfig.ExecStart = "/run/current-system/sw/bin/systemctl try-restart wpa_supplicant.service";
-  };
+  # Issue 3: Restart wpa_supplicant (and thereby connman) after rfkill unblock of wlan
+  #          This addresses the problem of wpa_supplicant with connman not seeing any
+  #          networks if wlan was initially soft blocked. (https://01.org/jira/browse/CM-670)
+  services.udev.packages = [ pkgs.rfkill_udev ];
+  environment.etc = [
+    { source = pkgs.writeScript "rfkill.hook" ''
+      #!${pkgs.runtimeShell}
+      if [ "$RFKILL_STATE" == 1 ]; then
```
guyonvarch: > I found out that the "normal" RFKILL_STATE is "1", while "0" is software-blocked, and "2" is hardware-blocked.

Could we add a comment indicating that `RFKILL_STATE` == 1 means it’s normal ?

knuton:
> Does the modification intends to completely fix WIFI setup ? Or is it a least an improvement on the situation ?

Because of the heuristic sleep, it might still fail in some cases. I tried a sleep duration of just 3 seconds, and this resulted in wifi working only some of the time. With 5 seconds, I have never observed non-functional wifi (but number of experiments is low). How much variation there is on different hardware for example, I don't know.

I consider this still a best-effort fix that seems to improve overall behavior and might be good enough in practice.

PR #30, opened 2020-01-07
Make the kiosk-browser easy to run standalone

guyonvarch:
Moreover, use the same packages as with the globally built version.


pkgs/default.nix
```diff
@@ -1,25 +1,46 @@
...
+    rev = "f52505fac8c82716872a616c501ad9eff188f97f";
+    sha256 = "0q2m2qhyga9yq29yz90ywgjbn9hdahs7i8wwlq7b55rdbyiwa5dy";
   };
 
-  playos-kiosk-browser = self.callPackage ../kiosk {
-    inherit system_version;
-    system_name = "PlayOS";
-  };
+  overlay =
+    self: super: {
```
guyonvarch: I adapted the usage of `self` and `super` according to [this section](https://nixos.org/nixpkgs/manual/#sec-overlays-definition) of the manual.

knuton:
:+1:

PR #31, opened 2020-01-10
Show PlayOS changelog in controller

guyonvarch:
Transform the changelog to HTML at build time with [Discount](https://www.pell.portland.or.us/~orc/Code/discount/) and show it with the same style as in Manager or Play.

This also fixes the content area that should is limited in both width and height, see [this issue](https://github.com/w3c/csswg-drafts/issues/1777).


guyonvarch:
I just tested it on a physical machine, and it works as expected.

knuton:
Would there be any downsides to generating the HTML at build time? It would reduce the runtime dependencies and might simplify the code too.

guyonvarch:
> Would there be any downsides to generating the HTML at build time? It would reduce the runtime dependencies and might simplify the code too.

The code would be only slightly simpler, we would remove those 2 lines:

```ocaml
|> Omd.of_string
|> Omd.to_html
```

The `omd` runtime dependency would be removed, but we would need `pandoc` for example at build time, that is not a small dependency.

In order to make it works, we would have to generate `Changelog.html` from `Changelog.md` in both `controller/default.nix`, and also during the developments, inside a helper command.

I am undecided on this, do you have a preference ?

knuton:
> The `omd` runtime dependency would be removed, but we would need `pandoc` for example at build time, that is not a small dependency.

Having less runtime dependencies is desirable both from a security standpoint as well as keeping download size low. `omd` is not much of a dependency, but it also doesn't seem likely to be needed at runtime.

As for build tools, I think if the user machines can take it, the dev/build machines can take it. There are very minimal implementations like [Discount](https://www.pell.portland.or.us/~orc/Code/discount/), which I think would suffice.

> In order to make it works, we would have to generate `Changelog.html` from `Changelog.md` in both `controller/default.nix`, and also during the developments, inside a helper command.

Not sure I understand the point about the helper command. I think the HTML could be generated and made available in a way similar to how the [PlayOS version info is filled in](https://github.com/dividat/playos/blob/6caefe03e2a51f13c08d76cef0b09e1a378616b9/controller/default.nix#L23-L29) (either as a string inside an Ocaml module or as a standalone HTML file).

By the way, it might be better to create a symlink to `Changelog.md` in the `controller` folder, rather than moving it.

guyonvarch:
> As for build tools, I think if the user machines can take it, the dev/build machines can take it. There are very minimal implementations like Discount, which I think would suffice.

`Discount` is a small and does the job, that’s all good. How did you find it ?

> Not sure I understand the point about the helper command. I think the HTML could be generated and made available in a way similar to how the PlayOS version info is filled in (either as a string inside an Ocaml module or as a standalone HTML file).

With the version, it’s not very problematic if it’s not replaced. But with the changelog, having a blank page prevent the developer to work on the page if it’s only replace with `nix-build`.

> By the way, it might be better to create a symlink to Changelog.md in the controller folder, rather than moving it.

A symlink does not work because `nix` build the controller in a isolated environment. If the symlink is relative, it fails. If it’s absolute, it misses permissions. It works with a hardlink, but it may not be ideal.

guyonvarch:
But the symlink would work the other way around, I’ll set this solution.

knuton:
:+1:

PR #32, opened 2020-01-15
Use common nixpkgs from controller

guyonvarch:
PlayOS is currently using the 19.03 version of nixpkgs. It contains
everything that was previously required for ocaml.


knuton:
Good move!

PR #33, opened 2020-03-12
Add persistence of preferred boot order

knuton:
The preferred boot order is persisted to preserve the boot choice after manual selection from the GRUB menu. By relying on the boot order, the manual choice will automatically be overridden following a successful update of the inactive partition, and fallback booting continues to work if the favoured choice fails repeatedly.

Fulfills two criteria:

- When manually selecting a system in the GRUB menu, the choice should be preserved.
- The manual boot choice should be overridden when a new update is found and installed.

### DIFF
--- a/bootloader/grub.cfg
+++ b/bootloader/grub.cfg
@@ -67,11 +67,15 @@ search --label system.b --set b_DISK
 set common_kernel_args="$machine_id_arg quiet panic=3 boot.panic_on_fail"
 
 menuentry "system.a (try=$a_TRY, ok=$a_OK)" --id system.a {
+  set ORDER="a b"
+  save_env ORDER
   linux ($a_DISK)/kernel root=/dev/disk/by-label/system.a rauc.slot=a $common_kernel_args
   initrd ($a_DISK)/initrd
 }
 
 menuentry "system.b (try=$b_TRY, ok=$b_OK)" --id system.b {
+  set ORDER="b a"
+  save_env ORDER
   linux ($b_DISK)/kernel root=/dev/disk/by-label/system.b rauc.slot=b $common_kernel_args
   initrd ($b_DISK)/initrd
 }


knuton:
Testing can not be done in the dev VM, as this does not set up a redundant A/B system. Build the installer ISO and install in a VM or on real hardware.

Open the GRUB menu like described here: https://github.com/dividat/playos/tree/develop/docs/user-manual#wiping-user-data

This should be enough to verify that the criteria above are met.

guyonvarch:
I am not able to run the `PlayOS install ISO`, it gets stuck.



Following the same procedure with `PlayOS live ISO`, I am able to run it.

I used the `nix build` command to produce the ISOs.

Do you have an idea what could go wrong ?

knuton:
Ah yes, sorry.

To make this work in VirtualBox, you want to change the graphics controller (Settings > Display) to VBoxSVGA (because of a NixOS issue: https://discourse.nixos.org/t/trying-to-fix-very-poor-virtualbox-install-experience/2488). You also need to assign a big enough hard disk. Dynamically allocated with maximum 60 GB should do.

Maybe we should document this somewhere?

guyonvarch:
> Ah yes, sorry.
> 
> To make this work in VirtualBox, you want to change the graphics controller (Settings > Display) to VBoxSVGA (because of a NixOS issue: https://discourse.nixos.org/t/trying-to-fix-very-poor-virtualbox-install-experience/2488). You also need to assign a big enough hard disk. Dynamically allocated with maximum 60 GB should do.
> 
> Maybe we should document this somewhere?

Thanks. Yes, we should document this, I can take care of this.

The display issue will disappear with Nhttps://github.com/NixOS/nixpkgs/commit/58d0134da072548eb66d9313ad629e4dffddfd9dixOS 20.03:

I still have a problem after the installation has been completed, I am not able to boot from the hard drive. Pressing `1)` bellow:



Here is my storage configuration:

knuton:
> I still have a problem after the installation has been completed, I am not able to boot from the hard drive.

Enable EFI in `Settings > System > Motherboard`.

guyonvarch:
> Enable EFI in `Settings > System > Motherboard`.

Thank you! That’s running now.

> Fulfills two criteria:
> 
> - When manually selecting a system in the GRUB menu, the choice should be preserved.
> - The manual boot choice should be overridden when a new update is found and installed.

I verified the first point. But concerning the second point, is it testable ?

knuton:
> But concerning the second point, is it testable ?

Yes, easiest way for a practical test is dropping the version number on this branch (e.g. to `2020.0.0`) and building e.g. a `master` channel build. Manual boot choice, then leaving system running until update is complete. Reboot and observe effects.

knuton:
> If the old system has been selected on purpose, we should not tell to reboot the computer. Maybe we miss another state, or we should just replace `RebootRequired` by `RunningOnOldVersion` ?

Good observation.

To provide a better status message I think we may want to get the primary slot from `rauc`. For this we need the change from https://github.com/rauc/rauc/commit/53fa129ef01908a2f39ce228731efe809d114a08, which is in versions `1.1` and up.

I will create a new ticket to upgrade `rauc` and subsequently improve the status.

PR #35, opened 2020-03-23
Document installation on VirtualBox

guyonvarch:
See the trello card: https://trello.com/c/uKTMzzdN/648-document-playos-installation-on-virtualbox

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -13,6 +13,8 @@ Running `nix build` will create following (in `result/`):
 - `playos-installer-VERSION.iso`: Bootable ISO image that can install the system.
 - `disk.img`: Preinstalled disk with bootloader, system partitions A/B and data partitions for testing (but without test instrumentation).
 
+### Choose what to build
+
 For quicker development cycles you may pass following arguments to the build:
 
 - `buildInstaller`: Should the installer ISO image be built.
@@ -55,15 +57,8 @@ When switching key pairs on a channel, the new certficiate must be built into th
 
 ## Installation on VirtualBox
 
-1. Build the installer with:
-
-```sh
-nix-build \
-  --arg buildInstaller true \
-  --arg buildBundle false \
-  --arg buildLive false \
-  --arg buildDisk false
-```
+1. Use `nix-build`. At least `buildInstaller` must be enabled. See
+   https://github.com/dividat/playos/tree/develop#choose-what-to-build.
 
 2. On VirtualBox, create a new virtual machine with:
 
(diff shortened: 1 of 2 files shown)


Readme.md
```diff
@@ -56,6 +53,36 @@ make develop
...
+
+1. Build the installer with:
+
+```sh
+nix-build \
+  --arg buildInstaller true \
+  --arg buildBundle false \
+  --arg buildLive false \
+  --arg buildDisk false
+```
```
knuton: Maybe a reference to the build arguments suffices for [self-help](https://en.wiktionary.org/wiki/give_a_man_a_fish_and_you_feed_him_for_a_day;_teach_a_man_to_fish_and_you_feed_him_for_a_lifetime)? https://github.com/dividat/playos/tree/develop#quick-start

Optionally adding a sub-heading to the arguments section to make it linkable might help.

Readme.md
```diff
@@ -56,6 +53,36 @@ make develop
...
+
+- `Settings > Display > Graphics controller:` set `VBoxSVGA` (see [this issue](https://discourse.nixos.org/t/trying-to-fix-very-poor-virtualbox-install-experience/2488), but it [should be resolved on NixOS 20.03](https://github.com/NixOS/nixpkgs/commit/58d0134da072548eb66d9313ad629e4dffddfd9d)),
+- `Settings > System > Motherboard`: enable EFI.
+
+3. Start the virtual machine. You will be prompted to select an optical drive,
+   select the previously built `result/playos-installer-VERSION.iso`. Then,
+   follow the installation procedure. Once the installation has been completed,
+   power off the virtual machine.
+
+4. In `Settings > Storage`, remove the optical drive containing
```
knuton: Points 3 and 4 are general points for installing most any Linux from an ISO.

guyonvarch:
> This is a good how-to, I just wonder whether it is too specific for the top-level Readme.
> 
> I left some comments with suggestions for focusing on the tricky bits. It also seems fine to leave the text as is, but I would consider moving it to a separate file and linking to it from the Readme then.

I agree, I moved it to `arch/Readme.md`, which has a section on testing on virtual machines. See [the live doc](https://github.com/guyonvarch/playos/tree/document-virtualbox-install/docs/arch#installation-on-virtualbox).

Readme.md
guyonvarch: I had some trouble being stuck rebooting on the ISO. I proposed a simplified version on those steps and with this indication.

Note: I am not used to install systems with VirtualBox.

knuton:
> I agree, I moved it to `arch/Readme.md`, which has a section on testing on virtual machines. See [the live doc](https://github.com/guyonvarch/playos/tree/document-virtualbox-install/docs/arch#installation-on-virtualbox).

Yeah, good call.

Small thing: The list of steps is disrupted, causing it to render as 1-2-1 instead of 1-2-3. Maybe you can fix that.

guyonvarch:
> Small thing: The list of steps is disrupted, causing it to render as 1-2-1 instead of 1-2-3. Maybe you can fix that. 

That’s fixed!

PR #38, opened 2020-03-26
Improve update status

knuton:
Adds update states `ReinstallationRequired` (when both systems have repeatedly failed) and `OutOfDateVersionSelected` (when the inactive system is newer, but the booted system is primary by being set as the preferred system in GRUB).

There is one remaining blip, which is that after the system health is updated, the update status does not immediately refresh. Because of this, `ReinstallationRequired` can continue to be displayed after the health and active system was already marked good. Fixing this would require changes to the health state to trigger an immediate refresh of the update state.

### DIFF
--- a/controller/server/update.ml
+++ b/controller/server/update.ml
@@ -166,7 +166,7 @@ let rec run ~update_url ~rauc ~set_state =
                 (* Inactive is up to date while booted is out of date, but booted was specifically selected for boot *)
                 OutOfDateVersionSelected |> set
               else
-                (* If booted is not up to date but inactive is up to date, we probably should reboot *)
+                (* If booted is not up to date but inactive is both up to date and primary, we should reboot into the primary *)
                 RebootRequired |> set
           | None ->
             (* All systems bad; suggest reinstallation *)
--- a/controller/server/update.ml
+++ b/controller/server/update.ml
@@ -119,7 +119,12 @@ type state =
   | ErrorDownloading of string
   | Installing of string
   | ErrorInstalling of string
+  (* inactive system has been updated and reboot is required to boot into updated system *)
   | RebootRequired
+  (* inactive system is up to date, but current system was selected for boot *)
+  | OutOfDateVersionSelected
+  (* there are no known-good systems and a reinstall is recommended *)
+  | ReinstallRequired
 [@@deriving sexp]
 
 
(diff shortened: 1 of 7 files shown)


controller/server/update.ml
```diff
@@ -149,16 +154,23 @@ let rec run ~update_url ~rauc ~set_state =
...
+              (* Don't care if inactive can be updated. I.e. Only update the inactive partition once the booted partition is outdated. This results in always two versions being available on the machine. *)
+              UpToDate version_info |> set
+            else
+              let%lwt booted_slot = Rauc.get_booted_slot rauc in
+              if booted_slot == primary_slot then
+                (* Inactive is up to date while booted is out of date, but booted was specifically selected for boot *)
+                OutOfDateVersionSelected |> set
+              else
+                (* If booted is not up to date but inactive is up to date, we probably should reboot *)
+                RebootRequired |> set
```
guyonvarch: The comment could be more precise:

```ocaml
(* Inactive is up to date while booted is out of date, and booted was not specifically selected for boot *)
```

Though, is it possible that booted was not specifically selected for boot, i.e. `booted_slot != primary_slot` ?

knuton: You're right, I left the comment that was there before, but it's not 100 % adequate now. I will update it.

`booted_slot != primary_slot` at this moment can only happen after an update has just been applied. Then e.g. system A is still booted, but system B has been made primary.

guyonvarch:
I made some tests and PlayOS rightly showed me `OutOfDateVersionSelected`. :+1: 

On `ReinstallationRequired`, if both systems (a and b) are mark as bad (try=3):

- if a run the old one, it shows `ReinstallRequired`,
- but If I run the up to date one, it shows `UpToDate`.

It seems `Rauc.get_primary` is returning something in case the up to date one is ran. The comment on top of `ReinstallRequired` becomes misleading.

Otherwise, we can add a Changelog entry for the current change.

guyonvarch:
:bulb: Idea: Show current installed versions next to system names.



```
*system.a 2020.1.0 (try=3, ok=1)
 system.b 2019.3.0 (try=3, ok=1)
```

Is it doable ?

knuton:
>  Idea: Show current installed versions next to system names.

Yes, I also had the thought that it would be handy. I think the way to implement it would be writing the version to a variable in the GRUB env.

knuton:
> On `ReinstallationRequired`, if both systems (a and b) are mark as bad (try=3):
> 
> * if a run the old one, it shows `ReinstallRequired`,
> * but If I run the up to date one, it shows `UpToDate`.

Is it possible that the up to date one showing `UpToDate` is the current `master` version, `2020.1.0`? That doesn't have the new code, so it would still show the old behavior.

PR #40, opened 2020-04-28
Update to nixpkgs 20.03

guyonvarch:
Nix derivation modifications:

- Rename `networking.connman` to `services.connman`
- Remove write permissions to breeze theme downloaded files
- Use `types.str` instead of `types.string`
- Adapt `environment.etc` file creation
- Rename `desktopManager.default` to `displayManager.defaultSession`

Controller:

- Remove client code that just did a test log (I had trouble building cohttp-lwt-jsoo)
- Use `ocaml-community/obus` instead of `dividat/obus`. I had an error
with `dividat/obus`, that’s the reason I investigated to use the
community version instead.

Kiosk:

- Add pyqtwebengine dependency
- Add missing `updateUrl` and `kioskUrl` to be able to run `nix-shell`
for developers

The important change is on the `obus` library. `dividat/obus` is forked from `steinuil/obus`, and we have a commit on top of it to use `dune`. Do you know why we needed this fork from `steinuil` ?

## Tests

On a physical machine, I verified that I could use a gamepad (the recorded version comes from a test on VirtualBox):

(website: http://gamepadviewer.com/)

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/system/modules/system-partition.nix
+++ b/system/modules/system-partition.nix
@@ -53,5 +53,15 @@ in
       ln -s mnt/system/init init
       cd /
     '';
+
+    # Replace /dev/root by /dev/disk/by-label/system.x in /etc/fstab
+    boot.postBootCommands = with pkgs; ''
+      SYSTEM_DEVICE=$(readlink /dev/root)
+      mv /etc/fstab /etc/fstab.old
+      cp -L /etc/fstab.old /etc/fstab
+      chmod u+w /etc/fstab
+      ${gnused}/bin/sed -i "s|/dev/root|$SYSTEM_DEVICE|" /etc/fstab
+      chmod u-w /etc/fstab
+    '';
   };
 }
--- a/system/modules/system-partition.nix
+++ b/system/modules/system-partition.nix
@@ -11,14 +11,14 @@ in
       device = mkOption {
         default = null;
         example = "/dev/sda";
-        type = types.string;
+        type = types.str;
         description = "Location of the device.";
       };
 
       fsType = mkOption {
         default = "auto";
         example = "ext3";
-        type = types.string;
+        type = types.str;
         description = "Type of the file system.";
       };
 
@@ -26,7 +26,7 @@ in
         default = [ "ro" ];
         example = [ "data=journal" ];
         description = "Options used to mount the file system.";
-        type = types.listOf types.string;
+        type = types.listOf types.str;
       };
 
     };
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,9 @@
 # Blatant misuse of make to create nice shortcuts for different build profiles. The real build tool is nix.
 
+# Allow unfree packages, required for zerotier using a BSL 1.1 licence
+# See https://nixos.wiki/wiki/FAQ/How_can_I_install_a_proprietary_or_unfree_package%3F
+export NIXPKGS_ALLOW_UNFREE=1
+
 # Get the git branch
 BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
 
--- a/Readme.md
+++ b/Readme.md
@@ -34,6 +34,9 @@ A helper is available to quickly start a virtual machine:
 make vm && ./result/bin/run-playos-in-vm
 ```
 
+In order to get the vm system journal, look at the output of `run-playos-in-vm`
+for a command starting with `socat`.
+
 See the output of `run-playos-in-vm --help` for more information.
 
 ## Deployment
--- a/bootloader/rescue/default.nix
+++ b/bootloader/rescue/default.nix
@@ -81,21 +81,21 @@ in
     };
 
     # Use ConnMan
+    services.connman = {
+      enable = true;
+      enableVPN = false;
+      networkInterfaceBlacklist = [ "vmnet" "vboxnet" "virbr" "ifb" "ve" "zt" ];
+      extraConfig = ''
+        [General]
+        AllowHostnameUpdates=false
+        AllowDomainnameUpdates=false
+
+        # Disable calling home
+        EnableOnlineCheck=false
+      '';
+    };
     networking = {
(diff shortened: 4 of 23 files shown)


knuton:
The changes look good. :+1:  Practical test to follow.

`nix-build` refused to build because of unfree package `zerotierone` ([BSL](https://www.zerotier.com/pricing/)). Our usage of ZeroTier is acceptable under their free license, as PlayOS is itself open source. We will want to add a note to the README explaining the situation and [how to deal with it](https://nixos.wiki/wiki/FAQ/How_can_I_install_a_proprietary_or_unfree_package%3F), however. Is it OK to just add `NIXPKGS_ALLOW_UNFREE=1` to the Makefile?

> The important change is on the `obus` library. `dividat/obus` is forked from `steinuil/obus`, and we have a commit on top of it to use `dune`. Do you know why we needed this fork from `steinuil` ?

This was done for `dune` migration, so it seems correct to switch to the community fork (cf. https://discuss.ocaml.org/t/should-obus-be-moved-to-ocaml-community/3946).

knuton:
The connman config in the installer needs to be switched to the new style: https://github.com/dividat/playos/blob/f5b37b2723f8d5a06ca2f3b25b7f9a71272f4219/installer/configuration.nix#L40-L52

guyonvarch:
> The connman config in the installer needs to be switched to the new style:

Done!

> We will want to add a note to the README explaining the situation and how to deal with it, however. Is it OK to just add NIXPKGS_ALLOW_UNFREE=1 to the Makefile?

I would have preferred to enable unfree packages only for `ZeroTier`:

```nix
{
  allowUnfreePredicate = pkg: builtins.elem (lib.getName pkg) [
    "zerotier"
  ];
}
```

or for every package using the `bsl11` licence:

```nix
{
  whitelistedLicenses = with stdenv.lib.licenses; [ bsl11 ];
}
```

But it needs to be done in the user configuration file.

See https://nixos.org/nixpkgs/manual/#sec-allow-unfree

I think it’s ok to use allow unfree in the Makefile.

guyonvarch:
When you mention an install system, it’s PlayOS installed on a physical machine ? on a VM ? You installed it with the install ISO ? Or did you applied the upgrade on an older system ?

knuton:
> When you mention an install system, it’s PlayOS installed on a physical machine ? on a VM ? You installed it with the install ISO ? Or did you applied the upgrade on an older system ?

Yes, sorry, I mean a fresh install from USB stick to a physical machine (typical machine used with customers).

guyonvarch:
Thanks, I am able to reproduce it when using the installer ISO in VirtualBox.

guyonvarch:
Installed systems are now working, though we might want to find a better solution for `/dev/root`.

I am now looking at connection issues.

guyonvarch:
Connection issues should be back to what’s on develop, `rfkill.hook` was not called because it was not executable.

system/modules/system-partition.nix
```diff
@@ -53,5 +53,15 @@ in
       ln -s mnt/system/init init
       cd /
     '';
+
+    # Replace /dev/root by /dev/disk/by-label/system.x in /etc/fstab
+    boot.postBootCommands = with pkgs; ''
+      SYSTEM_DEVICE=$(readlink /dev/root)
+      mv /etc/fstab /etc/fstab.old
+      cp -L /etc/fstab.old /etc/fstab
```
knuton, line 61: Two questions:

1. Why `mv` and `cp` back, instead of just copying?
2. Is there any use in keeping the `.old` copy around?

system/modules/system-partition.nix
```diff
@@ -53,5 +53,15 @@ in
...
+      chmod u+w /etc/fstab
+      ${gnused}/bin/sed -i "s|/dev/root|$SYSTEM_DEVICE|" /etc/fstab
+      chmod u-w /etc/fstab
+    '';
```
knuton, line 65: Thanks for figuring out where things go wrong!

While this change resolves the symptoms, I am not sure it solves the underlying problem yet. I tried deleting the whole entry for `/dev/root` from `/etc/fstab` and that also leads to a functioning system.

I am also not quite sure the changelog entry from systemd is relevant. According to the docs, systemd will create dependencies on the mount units for entries from `fstab`:

https://github.com/systemd/systemd/blob/1e5d2d656420d0e755dbcf72aeba3c3aba54e956/man/systemd.mount.xml#L181-L188

I also don't yet understand what exactly fails, seeing that the `/dev/root` link is actually already established. It's all a bit mysterious.

system/modules/system-partition.nix
guyonvarch: `/etc/fstab` is a symlink to `/etc/static/fstab`, wich is itself a symlink to `/nix/store/hash-etc-fstab`. We’re first moving the `/etc/fstab` link so that we can copy its content to `/etc/fstab`.

Another solution: get fstab nix store path with `read -f /etc/fstab`, remove `/etc/fstab`, cp nix store fstab to `/etc/fstab`.

We can remove the`.old` file, though it has no harm as it is.

system/modules/system-partition.nix
knuton: I need to come up from this rabbit hole ... :carrot: 

The whole phenomenon seems rather strange. There is a possibly related issue on the systemd project (https://github.com/systemd/systemd/issues/3446), with a panoply of different underlying problems and workarounds.

More workaround-heavy discussion here: https://unix.stackexchange.com/questions/556819/systemd-time-out-waiting-for-device-dev-mapper-vg

This rabbit hole might run deep and maybe your workaround is a good move in the end.

knuton:
I meant to ask to remove the `.old` copy, in case someone is ever spelunking around to and gets confused by it. But in that case, it may just as well prove useful as evidence of the dynamic change at boot time, so let's keep it.

PR #41, opened 2020-07-17
Enable use of HTTPS for dist host

knuton:
Previously the module ocaml-tls used by cohttp had spotty support for some modern TLS variants and hence http://dist.dividat.com/... was referenced as the update base URL.

While this does not put the self-update mechanism at danger due to a signature check on update bundles, it requires users to allow PlayOS machines to contact dist.dividat.com via HTTP. Being able to configure HTTPS means users can restrict network traffic to HTTPS for tighter control.

The ocaml-tls module just received a major upgrade, adding support for TLS 1.3 and new cryptographic methods. This patch updates to the latest version 0.12.0 of ocaml-tls.

This requires using a slightly newer version of nixpkgs, after 20.03.

## Doubts

- Will this lock out customers whose sysadmins already enabled HTTP but not HTTPS for the dist host?
- Will this cause problems in several years, when a "sleeping beauty" machine wakes up from a long slumber and can no longer talk to the HTTPS host with its by then advanced TLS version?

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -3,7 +3,7 @@
 let
 
   nixpkgs = builtins.fetchGit {
-    name = "nixpkgs-20.03";
+    name = "nixpkgs-20.03-snapshot";
     url = "git@github.com:nixos/nixpkgs.git";
     rev = "3f690bfcd4adde6dd0733c2d9f8f4d61e09dfc60";
   };
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -5,8 +5,7 @@ let
   nixpkgs = builtins.fetchGit {
     name = "nixpkgs-20.03";
     url = "git@github.com:nixos/nixpkgs.git";
-    rev = "5272327b81ed355bbed5659b8d303cf2979b6953";
-    ref = "refs/tags/20.03";
+    rev = "3f690bfcd4adde6dd0733c2d9f8f4d61e09dfc60";
   };
 
   overlay =
--- a/pkgs/ocaml-modules/obus/default.nix
+++ b/pkgs/ocaml-modules/obus/default.nix
@@ -11,8 +11,8 @@ buildDunePackage rec {
   src = fetchFromGitHub {
     owner = "ocaml-community";
     repo = "obus";
-    rev = "59c93162a4d4fc239761874f83d489332844e7c7"; # 1.2.0
-    sha256 = "0qb42634dmx8g6rf6vv5js88d8vgbzz8646dsg638352yzgsi3a3";
+    rev = "8aaf3d4e5538e42a62ae206dcfc01d2b898e54dc"; # 1.2.2
+    sha256 = "145c9ir0a4ld054npq80q8974fangirmd4r7z0736qjva27raqr7";
   };
 
   buildInputs = [ ];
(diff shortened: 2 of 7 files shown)


pkgs/ocaml-modules/obus/default.nix
```diff
@@ -11,8 +11,8 @@ buildDunePackage rec {
   src = fetchFromGitHub {
     owner = "ocaml-community";
     repo = "obus";
-    rev = "59c93162a4d4fc239761874f83d489332844e7c7"; # 1.2.0
-    sha256 = "0qb42634dmx8g6rf6vv5js88d8vgbzz8646dsg638352yzgsi3a3";
+    rev = "8aaf3d4e5538e42a62ae206dcfc01d2b898e54dc"; # 1.2.2
+    sha256 = "145c9ir0a4ld054npq80q8974fangirmd4r7z0736qjva27raqr7";
```
guyonvarch, line 15: Was the `obus` migration required to make it work with the update that went along the new versions of `nixpkgs` ?

pkgs/default.nix
```diff
@@ -5,8 +5,7 @@ let
   nixpkgs = builtins.fetchGit {
     name = "nixpkgs-20.03";
     url = "git@github.com:nixos/nixpkgs.git";
-    rev = "5272327b81ed355bbed5659b8d303cf2979b6953";
-    ref = "refs/tags/20.03";
+    rev = "3f690bfcd4adde6dd0733c2d9f8f4d61e09dfc60";
```
guyonvarch, line 8: Indicate in a comment, or in the name, that we’re not on the exact `nixpkgs-20.03` version ?

guyonvarch:
I verified that the new dist is downloadable via HTTPS with a VirtualBox PlayOS installation.

pkgs/ocaml-modules/obus/default.nix
knuton: Yes, the previous version relied on `type_conv` which in turn relied on the no longer supported `camlp4`.

knuton:
> Do we have a way to inform, either us or them, that the upload is blocked ?

In principle yes, we could do either. I wouldn't add alerting that informs us. After all, it is within a user's choice to block automatic updates.

It's conceivable to add some kind of messages in case an installation could not be updated for quite a while. For example, a check could be added that compares the current version's release time with the current time and warns if it has been a year since the release.

> Would there currently be systems that are currently blocked because only HTTPS is allowed ?

In fact yes, there could be such systems. The FAQ for a while incorrectly listed only HTTPS and not HTTP for `dist.dividat.com` for whitelisting.

Regarding the doubt I noted about the "sleeping beauty" effect: A machine dormant for long enough to miss enough updates to be affected by disabling of old TLS versions is more likely to fail to update due to rotated signature keys.

All in all, it seems best to take this step towards HTTPS.

guyonvarch:
> It's conceivable to add some kind of messages in case an installation could not be updated for quite a while. For example, a check could be added that compares the current version's release time with the current time and warns if it has been a year since the release.

Good idea. :+1:

pkgs/default.nix
knuton: Added `-snapshot` suffix.

knuton:
@guyonvarch Any further remarks? I am inclined to merge this and treat the warning about out-dated versions as a separate concern.

PR #42, opened 2020-07-17
Make gateway optional in connman's IPv{4,6} types

knuton:
The Gateway field is not present for all connections, hence failing on its absence can wrongly suggest that no IP is configured for an interface at all, when really there is simply no gateway.

## Checklist

-   [x] Changelog updated
-   [x] Code documented


knuton:
Link-local connection with this patch:

image

Link-local connection without this patch:

image

PR #43, opened 2020-09-29
Format machine-id for readability

knuton:
Make machine-id friendlier for human consumption.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/server/info.ml
+++ b/controller/server/info.ml
@@ -46,9 +46,26 @@ let of_file f =
   |> Mustache.of_string
   |> return
 
+(** Format a 32-character string as groups of 4; leave other strings unchanged. *)
+let format_machine_id s =
+  if String.length s == 32 then
+    String.concat
+      "-"
+      [ String.sub s 0 4
+      ; String.sub s 4 4
+      ; String.sub s 8 4
+      ; String.sub s 12 4
+      ; String.sub s 16 4
+      ; String.sub s 20 4
+      ; String.sub s 24 4
+      ; String.sub s 28 4
+      ]
+  else
+    s
+
 let get () =
   let%lwt ic = "/etc/machine-id" |> Lwt_io.(open_file ~mode:Input) in
-  let%lwt machine_id = Lwt_io.read ic >|= String.trim in
+  let%lwt machine_id = Lwt_io.read ic >|= String.trim >|= format_machine_id in
   let%lwt zerotier_address =
     (match%lwt Zerotier.get_status () with
      | Ok status -> Some status.address |> return
--- a/controller/server/info.ml
+++ b/controller/server/info.ml
@@ -46,26 +46,21 @@ let of_file f =
   |> Mustache.of_string
   |> return
 
-(** Format a 32-character string as groups of 4; leave other strings unchanged. *)
-let format_machine_id s =
-  if String.length s == 32 then
-    String.concat
-      "-"
-      [ String.sub s 0 4
-      ; String.sub s 4 4
-      ; String.sub s 8 4
-      ; String.sub s 12 4
-      ; String.sub s 16 4
-      ; String.sub s 20 4
-      ; String.sub s 24 4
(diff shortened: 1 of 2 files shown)


controller/server/info.ml
```diff
@@ -46,9 +46,26 @@ let of_file f =
...
+      "-"
+      [ String.sub s 0 4
+      ; String.sub s 4 4
+      ; String.sub s 8 4
+      ; String.sub s 12 4
+      ; String.sub s 16 4
+      ; String.sub s 20 4
+      ; String.sub s 24 4
+      ; String.sub s 28 4
+      ]
```
guyonvarch, line 62: While this works, it could be shorten using:

```ocaml
List.init 8 (fun x -> String.sub s (x * 4) 4)
```

Cons is that it may be less evident, but pro is that there is less potential error.

Another idea is to use a dedicated function:

```ocaml
let rec grouped n s =
  let l = String.length s in
  if l == 0 then
    []
  else if l <= n then
    [s]
  else
    List.cons (String.sub s 0 n) (grouped n (String.sub s n (l - n)))
```

The Ocaml std library does not contain a lot of functions. Scala is good on this:

```scala
scala> "1234ABCDZZZZAAAA1234ABCDZZZZAAAA".grouped(4).mkString("-")
res6: String = 1234-ABCD-ZZZZ-AAAA-1234-ABCD-ZZZZ-AAAA
```

controller/server/info.ml
```diff
@@ -46,9 +46,26 @@ let of_file f =
   |> Mustache.of_string
   |> return
 
+(** Format a 32-character string as groups of 4; leave other strings unchanged. *)
+let format_machine_id s =
+  if String.length s == 32 then
```
guyonvarch, line 51: According to `man machine-id`, a `machine-id` is exactly 32 characters long:

> The /etc/machine-id file contains the unique machine ID of the local system that is set during installation or boot. The machine ID is a single newline-terminated, hexadecimal, 32-character, lowercase ID. When decoded from hexadecimal, this corresponds to a 16-byte/128-bit value. This ID may not be all zeros.

guyonvarch (changes requested):

controller/server/info.ml
guyonvarch: So, I would be in favor to remove this condition.

controller/server/info.ml
knuton: Yes, I thought about this as well. The OCaml stdlib doesn't provide much and I agree that functions like those offered by Scala would be very useful here.

I am not so sure that introducing such generic code here is a win, however. While I _believe_ your function to be correct, I _know_ my code and your abbreviation (`List.init 8 (fun x -> String.sub s (x * 4) 4)`) to be correct. Our non-generic solutions are also about as short.

I think I'll go with your one-liner, which is probably clearest.

controller/server/info.ml
knuton: I recently dealt with a string formatting function deep in a project that assumed a certain shape and applied a specific transformation without verifying its assumptions, producing very garbled serial numbers when their shape changed.

So I think checking preconditions is a good thing, even if it is superfluous according to our current expectations.

knuton:
An antithesis to my comments and synthesis of your comments would be to drop the expectation of 32 characters and format in groups of 4 (then using `grouped`) irrespective of ID length. I think that would make sense, too.

guyonvarch:
> An antithesis to my comments and synthesis of your comments would be to drop the expectation of 32 characters and format in groups of 4 (then using grouped) irrespective of ID length. I think that would make sense, too.

Yes, this has the advantage of formatting malformed ids in a reasonable manner:

```scala
scala> "ABCDEFGHI".grouped(4).mkString("-")
res2: String = ABCD-EFGH-I
```

guyonvarch (approved):
Working as intended :+1:, tested in a VM.

PR #44, opened 2020-10-08
Allow to use proxies per connection in PlayOS

guyonvarch:
- Use provided hmap and opium from `nixpkgs`
- Run pacrunner daemon in the background
- Configure proxies from the controller
- Patch `ocaml-cohttp` to allows the utilization of proxies
- Use the defined proxy at startup in the controller and kiosk
- Show the currently used proxy in the controller

See [trello card](https://trello.com/c/6dp0jUuh/686-proxy-support-in-playos).

## UI

I initially thought of using a global proxy, but connman configures it per connection. Then, I also thought it was important to show the currently used proxy in the system. Do you think the current UI is too complicated ?

If we want a global proxy connection, we could bypass connman and use a `/var/gui` file.

## Solution to get proxy support in the controller

For the controller, `cohttp` unfortunately doesn’t support proxies. I got around 3 solutions:

- using curl with its `--proxy` parameter: if we want to get both the request status and the body outputs, we have to format the response adding the status code, and parse the all response.
- using the low level `callv` of cohttp (https://github.com/mirage/ocaml-cohttp/issues/459#issuecomment-565713210): it works, but I get this warning: `[WARNING] Closing channel failed: Unix.Unix_error(Unix.EBADF, "check_descriptor", "")`.
- apply a patch based on this comment (https://github.com/mirage/ocaml-cohttp/issues/459#issuecomment-186142690): it works, but I also get the same warning.

The current changeset implements the 3rd solution, but because we also get the same warning, it’s not better than solution 2, as I initially thought. And solution 2 does not involve patching `ocaml-cohttp`, hence should be more maintainable.

## Warning

In fact, we already got the warning I got with solutions 2 and 3, since the last `nixpkgs` update. We can look into it outside of this PR. I wondered if it was the same problem as https://github.com/mirage/ocaml-cohttp/issues/665, but probably not, since we are already using newer versions of libraries. I tried updating to `nixpkgs` 20.09 beta, and I still get the same warning.

## Testing

I am testing the change with a Rust proxy: https://gitlab.com/guyonvarch/proxy. Ideally, we should also only get internet access through the proxy.

## Checklist

-   [x] Changelog updated
-   [x] Code documented


controller/gui/template/network.mustache
```diff
@@ -36,6 +51,13 @@
 
             <pre class="d-Preformatted">{{properties}}</pre>
 
+
+            <form action="/network/{{id}}/proxy" method="post" class="d-Network__Form">
+              <legend class="d-Network__Legend">Proxy</legend>
+              <input type="text" class="d-Input d-Network__Input" value="{{proxy}}" name="proxy_url" placeholder="192.168.1.10:8500">
+              <input type="submit" class="d-Button" value="Validate">
+            </form>
```
knuton, line 59: Maybe this should be moved underneath the `connected` guard? It is a bit confusing to have two forms (with the proxy form appearing first) on a disconnected wifi. In addition, trying to configure a proxy on a disconnected wifi does not actually do anything, but claims to have been successful.

knuton (changes requested):
Nice! I can verify basic functionality in a local test setup.

I have a first batch of comments. Also, if it is not intricately connected to other changes in this PR I would suggest to create a separate PR for the dev server changes.

knuton:
> The current changeset implements the 3rd solution, but because we also get the same warning, it’s not better than solution 2, as I initially thought. And solution 2 does not involve patching `ocaml-cohttp`, hence should be more maintainable.

One thing to also watch out for is support for proxy authentication. Does the patched version of `ocaml-cohttp` support this? If not, it might be easier to just go with curl.

kiosk/kiosk_browser/proxy.py
```diff
@@ -0,0 +1,24 @@
+import subprocess
+import os
+from PyQt5.QtNetwork import QNetworkProxy
+
+def get_from_pacrunner():
+    return subprocess.run(["pacproxy"], capture_output=True, text=True).stdout.rstrip()
+
+def update_env(proxy):
+    os.environ['http_proxy'] = proxy
+    os.environ['https_proxy'] = proxy
```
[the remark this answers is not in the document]
guyonvarch: `use_in_qt_app` allows to use the proxy inside the QtWebEngine view, but requests made with the requests library is not affected. We could provide an explicit proxy, as done in https://github.com/dividat/playos/pull/44/commits/21c0d71c8a68ea0dac4c02f400d3f39a2a210253.

This question should be answered after the one questioning whether we should connect to the captive portal or the proxy first.

system/networking/pacproxy.py
```diff
@@ -0,0 +1,12 @@
+import dbus
+
+bus = dbus.SystemBus()
+
+client = dbus.Interface(
+    bus.get_object("org.pacrunner", "/org/pacrunner/client"),
+    "org.pacrunner.Client")
+
+result = "%s" % client.FindProxyForURL("https://dividat.com", "127.0.0.1")
```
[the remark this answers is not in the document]
guyonvarch: Documentation added with https://github.com/dividat/playos/pull/44/commits/a3ac933317ba12caa81f0e1f8eedab68bd11fd37.

controller/gui/template/network.mustache
guyonvarch: Done with https://github.com/dividat/playos/pull/44/commits/87aa57eb5f0916767de6825ac93570199426fdb1.

I just tested setting a proxy with an unconnected wifi:
- after having applied it, it does not show the proxy.
- but once we are connected to the network, it show it.

So, it works in the end, but the application is not giving an easy feedback.

Maybe we should only enable the proxy form to the connected network.

controller/gui/template/network.mustache
```diff
@@ -5,6 +5,21 @@
...
+        {{/is_proxy_set}}
+
+        {{^is_proxy_set}}
+            none
+        {{/is_proxy_set}}
+
+        <aside class="d-Network__Note">
+            Note that proxy change requires a restart.
+        </aside>
+    </dd>
```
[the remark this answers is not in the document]
guyonvarch: Done with https://github.com/dividat/playos/pull/44/commits/87aa57eb5f0916767de6825ac93570199426fdb1. I removed the top note, to put it in the modification place, but maybe a note in each place should be better.

controller/gui/template/network.mustache
guyonvarch: Proxy form opening with a toggle:

knuton: > Maybe we should only enable the proxy form to the connected network.

Yes, either that or fuse everything into one form to be submitted together (but the latter is more effort-full).

guyonvarch: Fusing the forms seems to be a really good idea, done in https://github.com/dividat/playos/pull/44/commits/69f6da0e2dced2cc8897f5e1c3308f29ba0c6278.



If we need the user to specify ipv4 or another field in the future, it can be extended easily.

Proxy information is still readable in the orange block.

guyonvarch:
> One thing to also watch out for is support for proxy authentication. Does the patched version of ocaml-cohttp support this? If not, it might be easier to just go with curl.

I set up a proxy on an external server using [squid](http://www.squid-cache.org/). Requiring a user and password to connect, I am able to use the proxy with firefox, but not using curl for the moment. Investigating.

guyonvarch:
I setup authentication using digest authentication:
https://wiki.squid-cache.org/Features/Authentication#Details

Using `--proxy-anyauth` should be working for `basic`, `ntlm`, `digest` and `negociate`.

There are some tweaks to use using curl with the controller, to get both the body and the status with curl, but that should be simpler compared to implementing ourselves different authentication schemes.

On the kiosk side, the proxy is used in QtWebEngine, and with requests to acces the captive portal. We may drop using the proxy for the captive portal, so I’ll concentrate on making it to work on QtWebEngine.

controller/gui/style.css
```diff
@@ -95,12 +100,24 @@
...
 
 .d-Network__Input {
     width: 50%;
-    margin-right: 1rem;
+    margin-bottom: 1rem;
+}
+
+.d-Network__Button {
+    width: 50%;
+    margin-top: 1rem;
```
knuton, line 120: Is this outcome intended?

image

controller/gui/template/network.mustache
```diff
@@ -34,20 +39,27 @@
...
 
-            {{^connected}}
-                <form action="/network/{{id}}/connect" method="post" class="d-Network__Form">
-                    <input type="text" class="d-Input d-Network__Input" name="passphrase">
-                    <input type="submit" class="d-Button" value="Connect">
-                </form>
-            {{/connected}}
+                        <input type="submit" class="d-Button d-Network__Button" value="Connect">
+                    </form>
+                {{/connected}}
```
knuton, line 55: Would it be hard to still allow for the proxy to be set on an already configured network? It's a bit awkward to have to remove the network in order to configure a proxy (and then have to retype the password).

controller/gui/template/network.mustache
```diff
@@ -34,20 +39,27 @@
...
+            <div class="d-Network__Properties">
+                <pre class="d-Preformatted">{{properties}}</pre>
+
+                {{^connected}}
+                    <form action="/network/{{id}}/connect" method="post" class="d-Network__Form">
+                        <label for="passphrase" class="d-Network__Label">Passphrase</label>
+                        <input type="text" class="d-Input d-Network__Input" id="passphrase" name="passphrase">
+
+                        <label for="proxy" class="d-Network__Label">Proxy</label>
+                        <input type="text" class="d-Input d-Network__Input" id="proxy" value="{{proxy}}" name="proxy" placeholder="192.168.1.10:8500">
```
knuton, line 51: Maybe this section could be put into a

```html
<details>
<summary>Advanced Settings</summary>
...
</details>

knuton (commented):
Yes, I think this direction is good. Some more feedback on the network form.

controller/gui/style.css
guyonvarch: Well, maybe better to reverted this: https://github.com/dividat/playos/pull/44/commits/ac6e4f1788c69cac311ac1c3f0fc5b006536cdfb.

controller/gui/template/network.mustache
guyonvarch: Tried with https://github.com/dividat/playos/pull/44/commits/ac6e4f1788c69cac311ac1c3f0fc5b006536cdfb.

controller/gui/template/network.mustache
guyonvarch: It’s a matter of adding back the previous method: https://github.com/dividat/playos/pull/44/commits/ac6e4f1788c69cac311ac1c3f0fc5b006536cdfb.

guyonvarch:
# Not connected with folded advanced settings



# Not connected with unfolded advanced settings



# Connected with folded advanced settings



# Connected with unfolded advanced settings

guyonvarch:
On proxy authentication, I am facing some difficulties:

- Querying `pacrunner` does not give the `user:proxy` if we save it in the url `user:proxy@host:port`. We could query first connman, and if it’s using a direct connection to a proxy, it’s storing `user:password`. Or we could store it in plain files as other configurations. There may also be security considerations to take into account.
- Setting up a distant or local proxy with [squid](http://www.squid-cache.org/), using `digest auth`, I am able to connect with `firefox`, `curl`, but not `qtwebengine`. Using `ncas basic auth`, I am able to connect with `firefox`, but not `curl` nor `qtwebengine`. We should be able to use basic auth with curl, I don’t know why it’s not working with `ncas` for the moment. It’s not even clear what `ncas` stands for. As for `qtwebengine`, there may be something to fix in order to use authentication.

knuton:
> * Setting up a distant or local proxy with [squid](http://www.squid-cache.org/), using `digest auth`, I am able to connect with `firefox`, `curl`, but not `qtwebengine`. Using `ncas basic auth`, I am able to connect with `firefox`, but not `curl` nor `qtwebengine`. We should be able to use basic auth with curl, I don’t know why it’s not working with `ncas` for the moment. It’s not even clear what `ncas` stands for. As for `qtwebengine`, there may be something to fix in order to use authentication.

This section in the Qt wiki looks relevant: https://wiki.qt.io/QtWebEngine/Network#Proxy_resolution_in_Qt_WebEngine

> It does only use the following properties though: QNetworkProxy::type(), QNetworkProxy::hostName(), QNetworkProxy::port()

> To authenticate proxies, connect to the QWebEnginePage::proxyAuthenticationRequired or WebEngineView::authenticationDialogRequested signal.

knuton:
> * Querying `pacrunner` does not give the `user:proxy` if we save it in the url `user:proxy@host:port`. We could query first connman, and if it’s using a direct connection to a proxy, it’s storing `user:password`.

It seems strange that `pacrunner` would not provide this information. Did you find any information on what _should_ be the case? Docs are sparse, but here's some mention of user/password in proxy URLs: https://kernel.googlesource.com/pub/scm/network/connman/pacrunner/+/0.8/doc/manual-configuration.txt#28

Can you explain the idea of querying connman in more detail?

guyonvarch:
> This section in the Qt wiki looks relevant: https://wiki.qt.io/QtWebEngine/Network#Proxy_resolution_in_Qt_WebEngine

Thanks! I think I got a working solution following this. But then, I had to use back the previous solution, setting a user and a password. It’s not completely clear what is going on, but at least, it’s working with basic authentication, with one way or another. I did not try the digest authentication.

A problem I am having is that I got a segfault when running the kiosk on my system. Updating to `nixpkgs-20.09` fix the problem. I tried updating only `qt`, or the `python` version, but none fixed the problem.

So, I started to have a look at upgrading `nixpkgs` to `20.09`, but I currently face an error building `gtk3`: `meson.build:516:97: ERROR: Dependency "cairo-xlib" not found, tried pkgconfig`.

`https://github.com/NixOS/nixpkgs/issues/102137`

> It seems strange that pacrunner would not provide this information. Did you find any information on what should be the case? Docs are sparse, but here's some mention of user/password in proxy URLs: https://kernel.googlesource.com/pub/scm/network/connman/pacrunner/+/0.8/doc/manual-configuration.txt#28

I found this discussion:
https://serverfault.com/questions/394130/how-to-automatically-configure-username-password-with-proxy-pac-file

> Can you explain the idea of querying connman in more detail?

We should be able to query connman first, through `dbus` as well, and if connman tells us that the proxy is configured manually, we can stop there and use the configured URL, that contains there user and password (tested with `cmst` for the moment).

guyonvarch:
Work is now done on https://github.com/dividat/playos/pull/54, closing this one.

PR #45, opened 2020-10-09
Add a captive portal simulator

guyonvarch:
It permits to try out captive portal integration.

## Checklist

-   ~~[ ] Changelog updated~~
-   [x] Code documented

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -58,17 +58,15 @@ make develop
 
 When switching key pairs on a channel, the new certficiate must be built into the bundle, which must then be signed with the old key. For this purpose, the `--override-cert` option of the deploy script is needed to provide RAUC with a certificate matching the new key.
 
-### Checklist
-
-- [ ] Update target branch (`master`/`validation`) with `git merge --no-ff --no-commit <source-branch>`
-- [ ] Set version in `default.nix`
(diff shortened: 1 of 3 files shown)


knuton (changes requested):
Can you add some more information in the dev-tools Readme about how this is intended to be used?

I suppose the idea is that the mock server's address is put in place of the regular captive portal endpoint? Is it intended for use when testing the kiosk in isolation or for an entire PlayOS system?

guyonvarch:
> Can you add some more information in the dev-tools Readme about how this is intended to be used?

https://github.com/dividat/playos/pull/45/commits/333a6668649e3654d33b4303ebdf4b2c1617f0ee

> I suppose the idea is that the mock server's address is put in place of the regular captive portal endpoint?

That’s it.

> Is it intended for use when testing the kiosk in isolation or for an entire PlayOS system?

I verified it worked with the kiosk in isolation, but it should also work for an entire PlayOS system if it has access to the simulated server.

The simulator can also be called from the controller, I also indicated this in https://github.com/dividat/playos/pull/45/commits/333a6668649e3654d33b4303ebdf4b2c1617f0ee.

knuton (approved):

PR #46, opened 2020-10-09
Use value instead of physical equality in controller

guyonvarch:
Because we are only interested in value equality, we should stick using
value (structural) equality rather than physical equality.

https://stackoverflow.com/questions/13590307/whats-the-difference-between-equal-and-identical-in-ocaml

From this small program:

```ocaml
let a = "foo" == "foo"
let b = "foo" = "foo"
let c = 1 == 1
let d = 1 = 1

let () = Printf.printf "a: %b\nb: %b\nc: %b\nd: %b\n" a b c d
```

It outputs:

```
a: false
b: true
c: true
d: true
```

I got bitten using the `==` equality on strings, which did not
work as I initially expected.

The current modification does not change the behavior of the program,
but it will help future changes not to use the more complex `==`
operator.

## Checklist

-   ~~[ ] Changelog updated~~
-   ~~[ ] Code documented~~

### DIFF
--- a/controller/bindings/rauc/rauc.ml
+++ b/controller/bindings/rauc/rauc.ml
@@ -58,7 +58,7 @@ let mark_good daemon slot =
       ("good", slot |> Slot.string_of_t)
   in
   let%lwt () = Logs_lwt.info ~src:log_src (fun m -> m "%s" msg) in
-  if Slot.of_string marked == slot then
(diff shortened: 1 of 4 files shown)


knuton (approved):
I also noticed this recently. Thanks! :+1:

PR #47, opened 2020-10-26
Auto reload the controller dev server

guyonvarch:
Previously, we had to manually stop and restart the server after
each change.

Dune accepts the `--watch` parameter, But it does not work with
`--exec`, see https://github.com/ocaml/dune/issues/2934.

There is a workaround, using an alias, see https://github.com/ocaml/dune/issues/2934#issuecomment-589087241.

`dune build --watch @install @run --profile release -j1 --no-buffer`

`server/dune`

```
(alias
  (name run)
  (action (run ./run-server)))
```

`run-server`

```bash
cd ../../..

function finish {
  # Stop the running server
  lsof -i tcp:3333 | grep LISTEN | awk '{print $2}' | xargs kill
}

trap finish EXIT

finish
./_build/install/default/bin/playos-controller
```

But there are 2 problems:

1. the server is started whenever the build succeeded or not. We could
find a solution to inquire whether the build ran successfully or not in
the `run-server` command.

2. A long output is appended after the error message, requiring
to scroll to see the error message. There may also be a solution to this
problem.

It seems simpler to stick to our custom solution at this point.

## Checklist

-   ~~[ ] Changelog updated~~
-   ~~[ ] Code documented~~

### DIFF
--- a/controller/bin/watch-command
+++ b/controller/bin/watch-command
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+tput reset
+
+markdown Changelog.md > Changelog.html
+
+# The controller application requires certain artifacts to be in a specific location relative to binary location. The `dune build @install` command ensures this.
+# The `--profile release` option disables dune from failing on warnings (which are currently present in the `obus` library)
+dune build @install --profile release
+
+if [ "$?" -eq 0 ]; then
+
+  # Kill already running server if any
+  lsof -i tcp:3333 | grep LISTEN | awk '{print $2}' | xargs kill 2>/dev/null
+
+  echo -e "$(tput setaf 2)$(tput bold)\nCompilation succeeded\n$(tput sgr0)"
+  ./_build/install/default/bin/playos-controller &
+  LIVE_SERVER_PID="$!"
+
+else
+
+  echo -e "$(tput setaf 1)$(tput bold)\nCompilation failed\n$(tput sgr0)"
+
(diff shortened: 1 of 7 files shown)


controller/bin/watch-command
```diff
@@ -0,0 +1,24 @@
...
+markdown Changelog.md > Changelog.html
+
+# The controller application requires certain artifacts to be in a specific location relative to binary location. The `dune build @install` command ensures this.
+# The `--profile release` option disables dune from failing on warnings (which are currently present in the `obus` library)
+dune build @install --profile release
+
+if [ "$?" -eq 0 ]; then
+
+  # Kill already running server if any
+  lsof -i tcp:3333 | grep LISTEN | awk '{print $2}' | xargs kill 2>/dev/null
```
knuton, line 14: Is the killing put behind this guard on purpose? It has the effect of letting the previously built server run if the new build fails, which I think could lead to confusion.

knuton (changes requested):
Works well, with one caveat.

controller/bin/watch-command
guyonvarch: Exact, fixed with https://github.com/dividat/playos/pull/47/commits/5c00e1e890e31f7cb6810acaebdd5e6ecec48af4. I also improved the initial launch of `bin/dev-server`, after a `dune clean`.

knuton (approved):

PR #49, opened 2020-11-23
Nixpkgs 20.09

guyonvarch:
## About removing `minimal.nix` import

This is done by https://github.com/dividat/playos/commit/a9669ca0470f1653ea6b7173eb250db2f39dd3f3.

The commit gives information on the change, but I just measured the time required to build the following command under different scenarii:

```
nix-build \
  --arg buildInstaller true \
  --arg buildBundle false \
  --arg buildLive false \
  --arg buildDisk false
```

Results:

- `develop`: 17 minutes
- `develop` without minimal: 6 minutse
- `nixpkgs-20.09` (this PR): 5 minutes

The additional time with `minimal.nix` is used compiling packages. Not importing `minimal.nix`, we are getting more (every ?) package from caches.

## Testing

- [x] qemu vm (I initially had an unworking connection, due to a wpa_supplicant error inside the vm, using `connman` with `wpa_supplicant` on the host machine. Switching to either network manager, or `connman` with `iwd` backend works however) 
- [x] live virtual box
- [x] installed virtual box

## Thoughts

I am testing `iwd` as a replacement to `wpa_supplicant` on my host machine. It may be a good candidate too in Play, we can wait and see if it’s performing well for me first.

## Checklist

-   [x] Changelog updated
-   ~~[ ] Code documented~~


knuton:
Nice finds, especially regarding the build times.

The system works in my tests, but I am observing some issues with error logs during startup (and in `journalctl` output):

1. [Time zone service](https://github.com/dividat/playos/blob/b8cbec830a0dce49b34795a17bb4b19142f0ec5a/system/localization.nix) is started incessantly for some reason (if TZ configured). Sometimes it succeeds, most of the time it fails.

```
Nov 26 14:08:05 playos systemd[1]: Failed to start Set system timezone.
Nov 26 14:08:05 playos systemd[1]: set-timezone.service: Start request repeated too quickly.
```

2. lvmconfig errors during second boot

From `dmesg`

```
[    1.553808] lvm2-activation-generator: lvmconfig failed
```

It may not cause practical issues, but looks unsettling. It seems that this is a general issue in 20.09 and a fix has landed upstream (yet to be merged into `20.09` channel): https://github.com/NixOS/nixpkgs/pull/103876/commits/8ac3a1503af31ed444b6997ff9eea781c2507348

Maybe we should just apply that patch in our project and add a reference to the issue so we can remove it later.

guyonvarch:
Thanks for your feedback.

> Time zone service is started incessantly for some reason (if TZ configured). Sometimes it succeeds, most of the time it fails.

I found a solution and added some information on the commit: https://github.com/dividat/playos/pull/49/commits/357efd8a6e2b451689af9965de0f288251b0d9c3.

I found out something that’s not working:

1. Set a timezone, like `Africa/Johannesburg`.
2. Set a shorter timezone, like `Africa/Accra`.
3. What is added to `/var/lib/gui-localization/timezone` is `Africa/Accranesburg`

But the problem is also happening on develop, so I’ll target develop on an external work.

> lvmconfig errors during second boot

I went with the patch, but it has to be applied 3 times:

- system,
- rescue,
- installer.

I am planning to also test the wifi on a live system on a media, I have some doubts looking at the logs about _wpa_supplicant_.

knuton:
> I am planning to also test the wifi on a live system on a media, I have some doubts looking at the logs about _wpa_supplicant_.

I tested it, though unsystematically, and found no issues. I can easily stress test it a bit more.

knuton:
> I went with the patch, but it has to be applied 3 times:

Maybe we could apply the patch once when [pinning nixpkgs](https://github.com/dividat/playos/blob/275bec0752648392efa54c3a1d397a0657b7e173/pkgs/default.nix#L5-L9) instead?

guyonvarch:
> Maybe we could apply the patch once when pinning nixpkgs instead?

Good idea, though I have trouble to find a way to do it.

We could pin to the commit that fixes the lvm2 warning, but then, we would also pull commits from master.

We could just take the same pin, `nixpkgs-20.09`, and apply the patch. It’s possible to apply a patch bellow the application of the overlay, but our patch concerns a nixos module, and I don’t know how to apply a patch to `importFromNixos`. Any idea ?

guyonvarch:
> I tested it, though unsystematically, and found no issues. I can easily stress test it a bit more.

It worked as expected on a live system running on a usb stick.

knuton:
> We could just take the same pin, `nixpkgs-20.09`, and apply the patch. It’s possible to apply a patch bellow the application of the overlay, but our patch concerns a nixos module, and I don’t know how to apply a patch to `importFromNixos`. Any idea ?

The example at the bottom here seems relevant, though a little involved: https://nixos.wiki/wiki/FAQ/Pinning_Nixpkgs

guyonvarch:
> The example at the bottom here seems relevant, though a little involved: https://nixos.wiki/wiki/FAQ/Pinning_Nixpkgs

Nice found, I was a little bit skeptical at first, but the result is cleaner than before: https://github.com/dividat/playos/pull/49/commits/a2abf8d88e0290b791c1fada6e7b674283ed753a.

knuton:
Thanks for the fixes!

- I verified that the lvm and timezone issues have been resolved.
- I also checked that the user-agent is still set as expected with the changed profile handling.

It would be nice to also fix the qemu connectivity issue. I think a fix is here: https://github.com/NixOS/nixpkgs/commit/8f177612b14063b644288a5a1058bf47f44b43a5

guyonvarch:
> It would be nice to also fix the qemu connectivity issue. I think a fix is here: NixOS/nixpkgs@8f17761

I am using back `wpa_supplicant` with `connman`, I had regularly a non working connection with `iwd`, after suspends if I remember correctly. I’ll test the patch. :+1:

I am wondering if this could also be fixing our issue with creating a dummy network, I will check further.

guyonvarch:
The fix does not fix the qemu connectivity.

```patch
commit 8f177612b14063b644288a5a1058bf47f44b43a5
Author: rnhmjoj <rnhmjoj@inventati.org>
Date:   Tue Nov 24 00:18:18 2020 +0100

    nixos/wireless: fix failure with no interfaces
    
    This resolves issue #101963.
    
    When the service is started and no interface is ready yet, wpa_supplicant
    is being exec'd with no `-i` flags, thus failing. Once the interfaces
    are ready, the udev rule would fire but wouldn't restart the unit because
    it wasn't currently running (see systemctl(1) try-restart).
    
    The solution is to exit (with a clear error message) but always restart
    wpa_supplicant when the interfaces are modified.

diff --git a/nixos/modules/services/networking/wpa_supplicant.nix b/nixos/modules/services/networking/wpa_supplicant.nix
index 39513987903..3cfcb535ef5 100644
--- a/nixos/modules/services/networking/wpa_supplicant.nix
+++ b/nixos/modules/services/networking/wpa_supplicant.nix
@@ -233,9 +233,10 @@ in {
       path = [ pkgs.wpa_supplicant ];
 
       script = ''
-        if [ -f /etc/wpa_supplicant.conf -a "/etc/wpa_supplicant.conf" != "${configFile}" ]
-        then echo >&2 "<3>/etc/wpa_supplicant.conf present but ignored. Generated ${configFile} is used instead."
+        if [ -f /etc/wpa_supplicant.conf -a "/etc/wpa_supplicant.conf" != "${configFile}" ]; then
+          echo >&2 "<3>/etc/wpa_supplicant.conf present but ignored. Generated ${configFile} is used instead."
         fi
+
         iface_args="-s -u -D${cfg.driver} -c ${configFile}"
         ${if ifaces == [] then ''
           for i in $(cd /sys/class/net && echo *); do
@@ -248,6 +249,10 @@ in {
               fi
             fi
           done
+          if [ -z "$args" ]; then
+            echo >&2 "<3>No wireless interfaces detected (yet)."
+            exit 1
+          fi
         '' else ''
           args="${concatMapStringsSep " -N " (i: "-i${i} $iface_args") ifaces}"
         ''}
@@ -261,7 +266,7 @@ in {
 
     # Restart wpa_supplicant when a wlan device appears or disappears.
     services.udev.extraRules = ''
-      ACTION=="add|remove", SUBSYSTEM=="net", ENV{DEVTYPE}=="wlan", RUN+="/run/current-system/systemd/bin/systemctl try-restart wpa_supplicant.service"
+      ACTION=="add|remove", SUBSYSTEM=="net", ENV{DEVTYPE}=="wlan", RUN+="/run/current-system/systemd/bin/systemctl restart wpa_supplicant.service"
     '';
   };
```

But I am not sure it help, connman seems to use ethernet with qemu, and does not seem to have any problem:



I verified that’s it’s no better if I am connected on ethernet on the host machine.

knuton:
After poking around a bit more, I realized that Internet connectivity _is_ completely functional, but DNS resolution is broken.

knuton:
I think this is a problem with the connman configuration in our host systems, which we probably sourced from the [ArchLinux wiki](https://wiki.archlinux.org/index.php/ConnMan#Blacklist_interfaces).

Using `tcpdump` inside the guest I could see that there were attempted DNS requests to `10.0.2.3`, which is [as it should be in qemu](https://www.qemu.org/docs/master/system/net.html#using-the-user-mode-network-stack).

I then probed this port (in the guest), `nmap -sU -p 53 10.0.2.3` and got the following:

    PORT   STATE         SERVICE
    53/udp filtered domain

So I tried turning off firewalls (`iptables`) in host and guest, but to no avail.

Finally I removed `"virbr"` from `networking.connman.networkInterfaceBlacklist` in my host's NixOS configuration, which restored DNS functionality inside the VM.

@guyonvarch Can you confirm?

guyonvarch:
> @guyonvarch Can you confirm?

Well done with the investigation!

I confirm that I have the same problem with `nmap`, and removing `virbr` from the connman blacklist works.

How do you understand that it was working before on `develop` ? Below the result of using `nmap` on `develop`, without removing the blacklist in the host configuration.

knuton:
> How do you understand that it was working before on `develop` ?

I don't. My guess would be that the user-space networking implementation in qemu was changed slightly and is now incompatible with connman ignoring the `virbr0` interface.

knuton (approved):
I am merging this and will add a note on setup for qemu.

PR #50, opened 2020-11-30
Show shortcuts for qemu VM

guyonvarch:
Being able to login as root with a shell permits to ease testing.

Moreover, we may forget from time to time shortcuts to get out of the
VM, or leave it.

Also take the opportunity to use a bash script instead of the Makefile.

Current `./build vm`:

```sh
+ nix-build --arg buildInstaller false --arg buildBundle false --arg buildLive false --arg buildDisk false
trace: warning: Using dummy update certificate. Build artifacts can only be used for local development.
/nix/store/vv8cvph6d7s015rixsivp1al31dm3z4x-playos-2020.7.0

Run ./result/bin/run-playos-in-vm to start a VM.

- unfocus: ctrl-alt-g
- quit: ctrl-alt-q
- tty: ctrl-alt-2, sendkey ctrl-alt-f1, ctrl-alt-1
```

Previous `make vm`:

```sh
nix-build \
	--arg buildInstaller false \
	--arg buildBundle false \
	--arg buildLive false \
	--arg buildDisk false
trace: warning: Using dummy update certificate. Build artifacts can only be used for local development.
/nix/store/vv8cvph6d7s015rixsivp1al31dm3z4x-playos-2020.7.0
Run ./result/bin/run-playos-in-vm to start a VM
```

## Checklist

-   ~~[ ] Changelog updated~~
-   ~~[ ] Code documented~~

PR #51, opened 2020-11-30
Fix localization state write

guyonvarch:
`Util.write_to_file` was writing the string to the file as a
replacement, not as a complete modification.

1. write Accra -> Accra (OK)
2. write Johannesburg -> Johannesburg (OK)
3. write Accra -> Accranesburg (KO)

Using the `O_TRUNC` flag truncate to 0 length if the file exists.
http://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#TYPEopen_flag

`Utils.write_to_file` is used on:

- /var/lib/gui-localization/timezone
- /var/lib/gui-localization/keymap
- /var/lib/gui-localization/lang

But only writes on the timezone file are affected, because keymaps and
langs all have the same length (see controller/server/gui.ml), hence the
bug does not appear there.

## Checklist

-   [x] Changelog updated
-   ~~[ ] Code documented~~

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -11,6 +11,7 @@
 ## Fixed
 
 - controller: Display interfaces' IP even if there is no gateway
+- controller: Fix timezone save when shorter than the previous saved one
 
 # [2020.7.0] - 2020-09-23
 
--- a/controller/bindings/util/util.ml
+++ b/controller/bindings/util/util.ml
@@ -20,7 +20,7 @@ let read_from_file log_src path =
 let write_to_file log_src path str =
   try
     let%lwt fd =
-      Lwt_unix.openfile path [ O_WRONLY; O_CREAT ] 0o755
+      Lwt_unix.openfile path [ O_WRONLY; O_CREAT; O_TRUNC ] 0o755
     in
     let%lwt bytes_written =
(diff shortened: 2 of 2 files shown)


guyonvarch:
## Testing

- Previously on the VM, I could not select a shorter timezone after having saved a longer one. (Check what’s happening to /var/lib/gui-localization/timezone)
- Now on the VM, I can select a shorter timezone after having saved a longer one. (Check what’s happening to /var/lib/gui-localization/timezone)

guyonvarch:
In order to inspect `/var/lib/gui-localization/`, it’s possible to connect to the tty of the VM, see https://github.com/dividat/playos/pull/50.

krksgbr (approved):
I tested the following way:

On `develop`:

1. Set timezone to Casablanca
```
cat /var/lib/gui-localization/timezone
Africa/Casablanca
```

2. Set timezone to Cairo
```
cat /var/lib/gui-localization/timezone
Africa/Cairolanca
```

On this branch:

1. Set timezone to Casablanca
```
cat /var/lib/gui-localization/timezone
Africa/Casablanca
```

2. Set timezone to Cairo
```
cat /var/lib/gui-localization/timezone
Africa/Cairo
```

I can also confirm that saving languages and keymaps still works fine.

PR #52, opened 2020-12-08
Improve qemu VM config and docs

knuton:
Followup to https://github.com/dividat/playos/pull/49#issuecomment-740194775.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/testing/run-playos-in-vm/run-playos-in-vm.py
+++ b/testing/run-playos-in-vm/run-playos-in-vm.py
@@ -54,7 +54,7 @@ def run_vm(system, qemu_opts, kernel_arguments):
             'rauc.slot=a', 'console=tty0', 'console=hvc1'
         ])
         initrd = sp + '/initrd'
-        virtfs_opts = 'local,path={},security_model=none,mount_tag=system,readonly'.format(
+        virtfs_opts = 'local,path={},security_model=none,multidevs=remap,mount_tag=system,readonly'.format(
             sp)
         print("\nsystem partition at:\n\t{}".format(sp))
         print("Kernel arguments:\n\t{}".format(kernel_arguments))
--- a/Readme.md
+++ b/Readme.md
@@ -39,6 +39,10 @@ for a command starting with `socat`.
 
 See the output of `run-playos-in-vm --help` for more information.
 
+#### Guest networking
+
+The default user-mode network stack is used to create a virtual Ethernet connection with bridged Internet access for the guest. If you find that the guest has a dysfunctional Internet connection, check your host's firewall settings. If using ConnMan, make sure that the virtual bridge interfaces (`virbr0` etc.) are not blacklisted.
+
 ## Deployment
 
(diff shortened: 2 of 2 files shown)


testing/run-playos-in-vm/run-playos-in-vm.py
```diff
@@ -54,7 +54,7 @@ def run_vm(system, qemu_opts, kernel_arguments):
             'rauc.slot=a', 'console=tty0', 'console=hvc1'
         ])
         initrd = sp + '/initrd'
-        virtfs_opts = 'local,path={},security_model=none,mount_tag=system,readonly'.format(
+        virtfs_opts = 'local,path={},security_model=none,multidevs=remap,mount_tag=system,readonly'.format(
```
guyonvarch, line 57: From the qemu manual:

> Specifies how to deal with multiple devices being shared with a 9p export. Supported behaviours are either "remap", "forbid" or "warn".  The latter is the default behaviour on which virtfs 9p expects only one device to be shared with the same export,  and if  more than one device is shared and accessed via the same 9p export then only a warning message is logged (once) by qemu  on host  side.  In  order to avoid file ID collisions on guest you should either create a separate virtfs export for  each  device to  be  shared  with  guests (recommended way) or you might use "remap" instead which allows you to share multiple devices with only  one  export  instead,  which is achieved by remapping the original inode numbers from host to guest in a way  that  would prevent  such collisions. Remapping inodes in such use cases is required because the original device IDs from  host  are  never passed  and  exposed  on  guest. Instead all files of an export shared with virtfs always share the same device id on guest. So two  files  with identical inode numbers but from actually dif‐ ferent devices on host would otherwise cause a file  ID  colli‐ sion  and  hence  potential misbehaviours on guest. "forbid" on the other hand assumes like "warn"  that  only  one  device  is shared by the same export, however it will not only log a warn‐ ing message but also  deny  access  to  additional  devices  on guest.  Note  though that "forbid" does currently not block all possible file access operations (e.g. readdir() would still re‐ turn entries from other devices).

Is it to share both the Senso and an Internet connection ?

knuton: I am actually not sure which devices it is complaining about, and we have not seen practical issues. I tried removing some device configurations but could not make the warning go away with default `multidevs` setting.

I am mostly trying to quell seemingly irrelevant noise. It's this kind of unhelpful error message that could easily be helpful by just telling you what the actual conflict is.

But I can also drop this commit, maybe it's unwise to hide this information (whatever it is pointing to).

guyonvarch: Ok, I forgot to see individual commits to get more information on this context before commenting. :sweat_smile: 

I don’t have a strong opinion on this, because of a lack of understanding. We can try this option.

guyonvarch (approved):

PR #54, opened 2020-12-10
Support proxies in PlayOS

guyonvarch:
## Review

I took care of keeping changes into different commits. The review should be easier looking at individual commits. That’s what I did yesterday to verify that the code was ready.

## Tests

With the latest version:

- System is using the configured proxy at startup, putting the proxy down is breaking the connection, and setting it up is making it up again. I applied an update through an unauthenticated proxy, I then restarted the system to the newly installed version (from develop).

With almost the latest version:

- download a system update through an authenticated (digest) proxy
- use an authenticated (basic) proxy
- label printer and zerotier requests are working

## Set up an authenticated proxy

Using squid with NixOS:

```nix
{
  services.squid = {
    enable = true;
    proxyPort = 1234;
    extraConfig = ''
      # Digest (generate credentials with: htdigest -c <path-to-pw> <realm> <username>)
      auth_param digest program ${pkgs.squid}/libexec/digest_file_auth -c <path>
      auth_param digest children 5
      auth_param digest realm <realm>

      # Basic (generate credentials with: htpasswd -c <path> <username>)
      # auth_param basic program ${pkgs.squid}/libexec/basic_ncsa_auth <path>

      acl users proxy_auth REQUIRED
      http_access allow users
    '';
  };
}
```

## Checklist

-   [x] Changelog updated
-   [x] Code documented


controller/bindings/label-printer/label_printer.ml
```diff
@@ -16,32 +16,18 @@ let json_of_label label =
   ]
```
guyonvarch, line 16: We could keep the old code using `cohttp` instead of `curl` for `label-printer` and `zetorier`, those changes are confined into a single commit.

controller/server/gui.ml
```diff
@@ -263,57 +261,125 @@ module NetworkGui = struct
...
+    in
+
+    let after_restart_proxy =
+      services
+      |> List.find_map (fun s ->
+          if s.Service.autoconnect then
+            Option.bind s.Service.proxy Proxy.validate
+          else
+            None)
+    in
```
guyonvarch, line 277: I tried to use [connman-proxy](https://github.com/dividat/playos/pull/54/files#diff-a00760ca559b79269f2bd1c6d1562867820c1e351358198ec6eb150d2d51f043) there, but I get a Unix pipe error. I was not able to get a more precise error, I wonder if it’s because of a dbus conflict.

controller/server/gui.ml
```diff
@@ -263,57 +261,125 @@ module NetworkGui = struct
...
+      [ proxy |> Option.map (Proxy.to_string ~hide_password:true)
+      ; (if after_restart_proxy <> proxy then
+          (match after_restart_proxy with
+          | Some p -> Some ("after restart: " ^ (Proxy.to_string ~hide_password:true p))
+          | None -> Some "after restart: removed")
+        else
+          None)
+      ]
+      |> Base.List.concat_map ~f:Option.to_list
+    in
```
guyonvarch, line 289: This permits to give useful insights to the user, indicating that his configuration is saved, but will only be applied after a restart.

controller/server/server.ml
```diff
@@ -52,8 +51,16 @@ let main debug port =
   (* Connect with ConnMan *)
   let%lwt connman = Connman.Manager.connect () in
 
+  let%lwt proxy = Connman_proxy.get () in
```
guyonvarch, line 54: Instead of relying on this external python script, we could get this with:

```ocaml
let%lwt proxy =
  Connman.Manager.get_services connman
  >|= List.find_map (fun s ->
    if s.Connman.Service.autoconnect then
      Option.bind s.Connman.Service.proxy Proxy.validate
    else
      None)
in
```

The advantage of using an external tool is to share the logic between the controller and the kiosk, but as a counterargument, we also have part of the logic duplicated in `gui.ml`.

system/networking/connman-manual-proxy.py
```diff
@@ -0,0 +1,35 @@
...
+def wait_for_connman():
+    # Waiting for connman, as it does not return the services until some
+    # seconds after startup (~4 seconds).
+    for i in range(0, 20):
+        if client.GetServices():
+            break
+        else:
+            time.sleep(0.5)
+
+wait_for_connman()
```
guyonvarch, line 22: If I remember correctly, this was not necessary when using `pacrunner`. I am not very happy with this solution.

controller/bindings/curl/curl.ml
```diff
@@ -0,0 +1,97 @@
...
+let safe_int_of_string str =
+  try
+    Some (int_of_string str)
+  with
+    Failure _ -> None
+
+let http_code_marker = '|'
+
+let parse_status_code_and_body str =
+  let open Base.Option in
```
guyonvarch, line 53: The standard library miss some usefull functions, which are covered by `base` from jane street:
https://ocaml.janestreet.com/ocaml-core/latest/doc/base/index.html

Discussion about using `base` or `batteries`:
https://www.reddit.com/r/ocaml/comments/azuqpr/base_vs_batteries/

knuton (commented):
The PR looks good overall!

I could verify basic functionality, but I am sometimes observing a lack of networks after booting. I want to check whether I can reproduce the same issues on `develop` (possibly an effect of #49).

In the meantime, here are some initial comments and suggestions for improvement.

controller/gui/template/network.mustache
```diff
@@ -34,20 +43,61 @@
...
+                                </form>
 
-            {{^connected}}
-                <form action="/network/{{id}}/connect" method="post" class="d-Network__Form">
-                    <input type="text" class="d-Input d-Network__Input" name="passphrase">
-                    <input type="submit" class="d-Button" value="Connect">
-                </form>
-            {{/connected}}
+                                <form action="/network/{{id}}/proxy/remove" method="post">
+                                    <input type="submit" class="d-Button" value="Remove">
```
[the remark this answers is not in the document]
guyonvarch: https://github.com/dividat/playos/pull/54/commits/e6b8e6abb30d312a9e6b276c6df82160eadcdf11

controller/server/gui.ml
```diff
@@ -263,57 +261,125 @@ module NetworkGui = struct
...
+          else
+            None)
+    in
+
+    let proxy_infos =
+      [ proxy |> Option.map (Proxy.to_string ~hide_password:true)
+      ; (if after_restart_proxy <> proxy then
+          (match after_restart_proxy with
+          | Some p -> Some ("after restart: " ^ (Proxy.to_string ~hide_password:true p))
+          | None -> Some "after restart: removed")
```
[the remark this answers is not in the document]
guyonvarch: I think the 2nd proposition can be ambiguous:



It’s not clear whether the new proxy setting is the one just above. We could add a hint just before the current proxy url: `Current: http://…`, but I am not sure if it’s enough. And we also lack a feedback from the update form, we would need to add an intermediate page saying: your proxy change for http:/… has been taken into account.

So, sticking to the wording improvement, I also applied the _note_ style:





Commit: https://github.com/dividat/playos/pull/54/commits/dbf7d62d7e22777b9953900bf7cd87056eae6d50

controller/server/gui.ml
```diff
@@ -263,57 +261,125 @@ module NetworkGui = struct
...
+      |> List.find_map (fun s ->
+          if s.Service.autoconnect then
+            Option.bind s.Service.proxy Proxy.validate
+          else
+            None)
+    in
+
+    let proxy_infos =
+      [ proxy |> Option.map (Proxy.to_string ~hide_password:true)
+      ; (if after_restart_proxy <> proxy then
```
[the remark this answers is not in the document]
guyonvarch: The proxy detection must be different between startup (done by the Python script) and during the request. In both cases, it’s looking for a service which has _AutoConnect_ set to _true_, and pick the 1st one.

Using ethernet, but no wifi, I don’t have the problem. Do you have both ethernet and wifi at the same time ?

 Can you reproduce this directly on the controller ? Or in the qemu VM ?

controller/gui/template/network.mustache
```diff
@@ -34,20 +43,61 @@
...
+                        <input type="text" class="d-Input d-Network__Input" id="{{id}}-passphrase" name="passphrase">
+
+                        <details>
+                          <summary>Advanced Settings</summary>
+
+                          <div class="d-Network__AdvancedSettingsTitle">
+                            <div class="d-Network__Label">
+                              <label for="{{id}}-proxy">Proxy</label>
+                            </div>
+                            <input type="text" id="{{id}}-proxy" class="d-Input d-Network__Input" name="proxy" placeholder="http://user:password@host:port">
```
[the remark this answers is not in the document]
guyonvarch: I took some inspiration from SmartTouch configuration form: https://github.com/dividat/playos/pull/54/commits/844f90d34619b23e10d6e0b78a19aed28c9d6369

controller/server/gui.ml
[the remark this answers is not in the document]
guyonvarch: This has changed a bit with https://github.com/dividat/playos/pull/54/commits/844f90d34619b23e10d6e0b78a19aed28c9d6369:

controller/server/gui.ml
```diff
@@ -263,57 +261,125 @@ module NetworkGui = struct
...
+      form_data
+      |> List.assoc "proxy"
+      |> List.hd
+    in
+    if String.trim proxy_str = "" then
+      return None
+    else
+      match Proxy.validate proxy_str with
+      | Some proxy -> return (Some proxy)
+      | None -> fail_with (Format.sprintf "Proxy '%s' is not valid." proxy_str)
```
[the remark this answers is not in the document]
guyonvarch: https://github.com/dividat/playos/pull/54/commits/844f90d34619b23e10d6e0b78a19aed28c9d6369#diff-e14b53e1ae4d8359d75e6687367901450847f7f724c8989d2ecf701e82526f11R331

controller/proxy/proxy.mli
```diff
@@ -0,0 +1,32 @@
...
+  ; host: string
+  ; port: int
+  }
+
+val validate : string -> t option
+(** [validate str] returns [t] if [str] is valid.
+
+    Valid proxies:
+
+      - Use the http scheme,
```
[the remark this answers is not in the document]
knuton: Thanks for explaining. According to this issue, there are no plans to support HTTPS proxies in Qt: https://bugreports.qt.io/browse/QTBUG-66601

It seems fine given the landscape to also explicitly limit support to HTTP proxies for now.

controller/server/gui.ml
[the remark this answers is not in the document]
knuton: :+1:

controller/server/gui.ml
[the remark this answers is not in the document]
knuton: > Using ethernet, but no wifi, I don’t have the problem. Do you have both ethernet and wifi at the same time ?

Only Ethernet is connected.

> Can you reproduce this directly on the controller ? Or in the qemu VM ?

I get this result when testing in a full PlayOS system installed on real hardware. I see `playos-controller: [INFO] proxy: none` in the logs.

Is it possible that the detection initially finds no proxy, since it runs before the Ethernet interface appears or the router on the other side activates?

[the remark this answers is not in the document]
guyonvarch: I am reproducing the problem on my system, I got insight by adding logs to the detection script:
https://github.com/guyonvarch/playos/commit/fbc46aed937455b5dac82a8530d7625d940531b1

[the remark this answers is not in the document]
knuton: Since we are now configuring the proxy inside the running apps and no longer using environment variables, might it be an idea to subscribe to connection changes via DBus and update proxy information accordingly?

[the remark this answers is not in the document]
guyonvarch: Yes, that would be less brittle, and it would be more valuable to users.

I just verified that we can change the proxy on the fly on the kiosk, it’s working well. So, I think there is nothing preventing us to set this up. I’ll have a try at this. :slightly_smiling_face:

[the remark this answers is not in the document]
guyonvarch: Live reload of the proxy in the kiosk: https://github.com/dividat/playos/pull/54/commits/41a69d4257ca5ae0b4ead11ec0e3a5602f67d626

I wrote about an unlikely, but still potential issue on initialization, I’ll think how to take care of it.

But I’ll take care of the controller first. My idea is not to listen to _connman_ for changes in the controller, but rather to get a fresh proxy everytime we need it.

guyonvarch:
I had the network down again in the _qemu vm_, and strangely, removing `  # services.connman.networkInterfaceBlacklist = [ "vmnet" "vboxnet" "ifb" "ve" ];` from my system configuration did the trick.

I don’t understand why the fix that permitted to make the network to work in the vm is now blocking it, without having done an upgrade of my system in the mean time.

controller/server/gui.ml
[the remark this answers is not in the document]
knuton: Nice, did a quick test of the current state and it seems to perform well!

> I wrote about an unlikely, but still potential issue on initialization, I’ll think how to take care of it.

Can the problem be avoided by actively polling once for the current state just after setting up the monitoring?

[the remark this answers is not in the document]
guyonvarch: I am having an issue on _qemu VM_: if a select a non existent proxy, the network page stills in the controller stills indicate I am connected. I’ll see what’s happening. I don’t have the problem running the controller directly on my system.

> Can the problem be avoided by actively polling once for the current state just after setting up the monitoring?

￼Yes, that should do it. I am not completely sure at which point we can be sure the monitoring is set up. I’ll have to understand better how the dbus main loop is working.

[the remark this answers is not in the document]
guyonvarch: > I am having an issue on qemu VM: if a select a non existent proxy, the network page in the controller stills indicate I am connected. I’ll see what’s happening. I don’t have the problem running the controller directly on my system.

Fixed with https://github.com/dividat/playos/pull/54/commits/baa96d0a4c0edc699809d3975f83b87f70eae779, which is recomputing the internet status on the fly, instead of using a computed state.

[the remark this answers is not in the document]
guyonvarch: > Can the problem be avoided by actively polling once for the current state just after setting up the monitoring?

Initializing the proxy after the monitoring is on: https://github.com/dividat/playos/pull/54/commits/6587316e832854245885c31bb2706cd3eacb2ef0

I can’t say if the monitoring is on directly after the call is made, but at least, the loop behind has nothing to do with it, I verified it by adding a sleep before the loop is call, and verifying proxy updates are detected and applied.

system/networking/connman-manual-proxy.py
guyonvarch: This code has been removed, in favor of monitoring proxy changes.

controller/server/server.ml
guyonvarch: We’re now getting the proxy directly in Ocaml, as proposed above.

controller/server/gui.ml
guyonvarch: No longer showed.

controller/server/gui.ml
guyonvarch: Not relevant anymore.

guyonvarch:
One last thing, I’ll set back confirmation messages after a proxy modification / removal. It’s not mandatory to understand that the change has taken place, because the information is shown in the network panel, but it’s more consistent with other network actions.

guyonvarch:
I checked that the VM is working as intended using different configurations:

- no proxy,
- local proxy not running,
- local proxy runnnig.

I think we’re not far from the end now!

guyonvarch:
I am improving proxy initialization a bit in the kiosk, setting to changes requested. We want to ensure the proxy is setup with the correct value before opening the first URL.

guyonvarch:
> I am improving proxy initialization a bit in the kiosk, setting to changes requested. We want to ensure the proxy is setup with the correct value before opening the first URL.

Done with https://github.com/dividat/playos/pull/54/commits/5694f80ebd8ea10f58ffd064bd52ebf0e0deb735.

I verified in the VM that I could still use an authenticated proxy.

kiosk/kiosk_browser/browser_widget.py
```diff
@@ -44,6 +54,15 @@ def _load_finished(self, success):
...
             QTimer.singleShot(5000, self.reload)
 
+    def _proxy_auth(self, get_current_proxy, url, auth, proxyHost):
+        proxy = get_current_proxy()
+        if proxy is not None and proxy.username is not None and proxy.password is not None:
+            logging.info("Authenticating proxy")
+            auth.setUser(proxy.username)
+            auth.setPassword(proxy.password)
+        else:
+            logging.info("Ignoring proxy authentication")
```
knuton, line 64: I am not sure I understand what this log says. Reading the conditional above I get the impression that no credentials were given, but this log reads like the kiosk somehow chose to ignore the credentials.

kiosk/kiosk_browser/browser_widget.py
```diff
@@ -1,21 +1,31 @@
...
 
 from kiosk_browser import system
 
 class BrowserWidget(QWebEngineView):
 
-    def __init__(self, url, *args, **kwargs):
+    def __init__(self, url, get_current_proxy, *args, **kwargs):
         QWebEngineView.__init__(self, *args, **kwargs)
 
+        # Authenticate proxies
```
knuton, line 16: Nitpick: Update comment to clarify that this merely registers the handler for authentication, should it be required.

kiosk/kiosk_browser/connection.py
```diff
@@ -6,20 +10,23 @@
 
 check_connection_url = 'http://captive.dividat.com/'
 sleep_time_not_connected = 5
-sleep_time_connected = 60
+sleep_time_connected = 6
+sleep_time_proxy = 6
```
knuton, line 14: These seem quite low to me. It's nice from a user's perspective, but sysadmins might find our device needlessly chatty when idle. Did you choose these durations with a specific scenario in mind?

kiosk/kiosk_browser/connection.py
```diff
@@ -6,20 +10,23 @@
...
 
 class Status(Enum):
     DISCONNECTED = auto()
     CAPTIVE = auto()
     CONNECTED = auto()
+    PROXY = auto()
```
knuton, line 20: I find the blurring between the status a bit confusing. When behind a proxy we can still be connected or disconnected. Would it make sense to prefix the non-proxy status with `DIRECT_` to hint at them being a "world of their own".

kiosk/kiosk_browser/main_widget.py
```diff
@@ -5,17 +5,23 @@
...
 
     def __init__(self, urls, toggle_sequence):
         super(MainWidget, self).__init__()
 
-        self._connection = connection.Connection(self.set_captive_portal_url)
+        proxy = proxy_module.Proxy()
+        proxy.start_monitoring_daemon()
+
+        self._connection = connection.Connection(proxy.get_current, self.set_captive_portal_url)
+        self._connection.start_monitoring_daemon()
```
knuton, line 19: Maybe this could be named more specifically now to reflect that it is meant to detect captive portals?

kiosk/kiosk_browser/proxy.py
```diff
@@ -0,0 +1,113 @@
...
+def get_current_proxy(bus):
+    """Get current proxy from dbus
+
+    Return  the proxy of a ready or connected service."""
+
+    client = dbus.Interface(
+        bus.get_object('net.connman', '/'),
+        'net.connman.Manager')
+
+    connected_services = [s for s in client.GetServices() if s[1]['State'] in ['ready', 'online']]
```
knuton, line 51: Will this make sure that `online` services are preferred? I think they should be.

Can we find the service picked as default gateway?

controller/proxy/proxy.ml
```diff
@@ -0,0 +1,56 @@
...
+  ; ":"
+  ; string_of_int t.port
+  ]
+  |> String.concat ""
+
+let from_connected_service services =
+  let open Connman.Service in
+  services
+  |> List.find_map (fun s ->
+    if s.state = Ready || s.state = Online then
```
knuton, line 53: Same remark as in Python version, it might be useful to prefer `Online` services/default gateway.

knuton (commented):
Looks good! Here are a bunch of code review comments, functional testing is next.

controller/Changelog.md
```diff
@@ -3,6 +3,7 @@
 ## Added
 
 - controller: Enable HTTPS support for system update hosts
+- controller: Support proxies
```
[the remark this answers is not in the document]
guyonvarch: https://github.com/dividat/playos/pull/54/commits/ed80c955e0dd26fbd1dbc839b0be46cd3ebdd94c

controller/gui/template/network.mustache
```diff
@@ -34,20 +39,67 @@
...
+                        <details>
+                          <summary>Advanced Settings</summary>
+
+                          <div class="d-Network__AdvancedSettingsTitle">
+                            <div class="d-Network__Label">
+                              <label for="{{id}}-proxy">Proxy</label>
+                            </div>
+                            <input type="text" id="{{id}}-proxy" class="d-Input d-Network__Input" name="proxy">
+                            <p class="d-Network__Note">
+                                URL in the form <em class="d-Code">http://user:password@host:port</em>
```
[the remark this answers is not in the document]
guyonvarch: https://github.com/dividat/playos/pull/54/commits/c9bf24ec8757dca61b729544ca0ea296b52fafd1

controller/server/gui.ml
```diff
@@ -245,75 +242,128 @@ module NetworkGui = struct
...
-    | Some service ->
-      Connman.Service.remove service
-      >|= (fun () -> Format.sprintf "Removed service %s." service.name)
-      >>= success
-
-  let build
-      ~(connman:Connman.Manager.t)
-      ~(internet:Network.Internet.state Lwt_react.S.t)
-      app =
+      fail_with "Proxy must not be empty."
```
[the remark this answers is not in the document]
guyonvarch: https://github.com/dividat/playos/pull/54/commits/c9bf24ec8757dca61b729544ca0ea296b52fafd1

controller/gui/template/network.mustache
```diff
@@ -34,20 +39,67 @@
...
 
-            {{^connected}}
-                <form action="/network/{{id}}/connect" method="post" class="d-Network__Form">
-                    <input type="text" class="d-Input d-Network__Input" name="passphrase">
-                    <input type="submit" class="d-Button" value="Connect">
-                </form>
-            {{/connected}}
+                                {{#has_proxy}}
+                                    <form action="/network/{{id}}/proxy/remove" method="post">
+                                        <input type="submit" class="d-Button" value="Remove">
```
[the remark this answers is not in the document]
guyonvarch: https://github.com/dividat/playos/pull/54/commits/c9bf24ec8757dca61b729544ca0ea296b52fafd1

kiosk/kiosk_browser/browser_widget.py
guyonvarch: Adapting the log in https://github.com/dividat/playos/pull/54/commits/811d5c06647486393f4cdf76aae3a041fecfb4ff

kiosk/kiosk_browser/browser_widget.py
guyonvarch: https://github.com/dividat/playos/pull/54/commits/811d5c06647486393f4cdf76aae3a041fecfb4ff

kiosk/kiosk_browser/connection.py
guyonvarch: I don’t remember, I may have put small numbers for testing purpose, forgetting to put it back to 60. Putting it back to 60 in idle cases in https://github.com/dividat/playos/pull/54/commits/082b207e80f2cf871decfdcf2576021e1b15e5fd and improving the mechanic, logging a bit.

kiosk/kiosk_browser/connection.py
guyonvarch: That’s right. Prefixing with `DIRECT_` and adding a comment should be enough, done in https://github.com/dividat/playos/pull/54/commits/e1293840cadf356ef5f45c0f53c68eb21e38b58a.

kiosk/kiosk_browser/main_widget.py
guyonvarch: https://github.com/dividat/playos/pull/54/commits/b010e485194c6581beea971050119b690a4b6a9f

kiosk/kiosk_browser/proxy.py
guyonvarch: I was wondering whether it was possible to have many services in `online` or `ready` state, it seems so according to `doc/manager-api.txt`:

> If at least one service is in "ready" state and no service is in "online" state, then it will indicate the "ready" state.

https://github.com/dividat/playos/pull/54/commits/eaf3ac2227e903ea4d71ec72de39d66f297677de

controller/proxy/proxy.ml
guyonvarch: https://github.com/dividat/playos/pull/54/commits/eaf3ac2227e903ea4d71ec72de39d66f297677de

controller/bindings/curl/curl.ml
```diff
@@ -0,0 +1,101 @@
...
+      Printf.sprintf "curl killed by signal %d" n
+
+  | ProcessStop n ->
+      Printf.sprintf "curl stopped by signal %d" n
+
+type result =
+  | RequestSuccess of int * string
+  | RequestFailure of error
+
+let exec cmd =
```
knuton, line 29: I think it might be prudent to put this whole function body into a try-with block. According to the [Lwt docs](https://ocsigen.org/lwt/4.5.0/api/Lwt_io):

> input functions of this module raise `End_of_file` when the end-of-file is reached (i.e. when the read function returns 0). Other exceptions are ones caused by the backend read/write functions, such as `Unix.Unix_error`. `exception Channel_closed of string` Exception raised when a channel is closed.

The calls to `Unix.pipe` can themselves raise exceptions.

So it seems we need to expect at least the following exceptions:

- `Unix.Unix_error`
-  `End_of_file`
- `Channel_closed`

A similar situation already exists here: https://github.com/dividat/playos/blob/9e9d243be2f027e58cf8b1a7b5e52150d0e1522e/controller/bindings/util/util.ml#L6-L16

knuton: Related: https://github.com/dividat/playos/pull/59

kiosk/kiosk_browser/proxy.py
knuton: It is indeed possible, and the situation is made worse by connman seemingly not using the proxy information it has to perform the online check.

I just tested and observed that a client whose MAC is blocked from direct Internet access in the router will remain in `ready`. I don't understand whether or not this is the intended behaviour (https://git.kernel.org/pub/scm/network/connman/connman.git/tree/src/wispr.c).

We may want to think about also preferring services that have a gateway configured (_online_ > _gateway_ > _ready_).

knuton: Looking at [this mailing list entry](https://lists.01.org/hyperkitty/list/connman@lists.01.org/message/GLKCXNUXPPUOARE5JURQXDG43RUVETXV/) it seems that connman itself relies on pacrunner to detect the proxy for a connection when making the online check: 

> If ConnMan gets to 'online' state, that is only a hint to the user that every connection should now
be working. Even the proxy ones - provided that the applications then is able to use for example pacrunner as there might still be a proxy preventing regular non-http PAC aware applications from functioning properly. Normally nobody runs pacrunner, meaning that if the host needs a proxy, the connection will still stay at 'ready'.

https://git.kernel.org/pub/scm/network/connman/connman.git/tree/src/proxy.c#n91
https://git.kernel.org/pub/scm/network/connman/connman.git/tree/plugins/pacrunner.c#n360

knuton:
Shelving final review until safe-guarding of `Unix` calls and proxy selection has been clarified.

controller/bindings/curl/curl.ml
guyonvarch: Thanks for your analysis!

Catching exceptions is done in https://github.com/dividat/playos/pull/54/commits/a5d32d007cd913edc56b7e6602890ceb1045242f.

Errors are integrated into curl error type, no logging is directly applied. I’ll check if callers are taking care of the error case.

I am not completely sure it’s necessary to catch exceptions outside of Lwt, with the same mechanism we introduced in request handlers.

controller/bindings/label-printer/label_printer.ml
knuton: No, good move to use the curl wrapper everywhere, I think.

controller/bindings/curl/curl.ml
guyonvarch: Logging request errors reaching the captive portal in https://github.com/dividat/playos/pull/54/commits/6d63f9c3b012fea7146b8c9bb7ee367c910766f0.

Logging is taken care of with requests done by `update`, but not for those done by `zerotier` and `label_printer` I think.

guyonvarch: Logging `zerotier` errors as well in https://github.com/dividat/playos/pull/54/commits/bddcbb0bc70f0094df6c5ddb4e6fc8fb4176d071.

For `label_printer`, it’s already taken care of by the request handler.

kiosk/kiosk_browser/proxy.py
guyonvarch: > Looking at this mailing list entry it seems that connman itself relies on pacrunner to detect the proxy for a connection when making the online check

If I understand it correctly: if we don’t run pacrunner, connections behind a proxy will remain in _ready_ state. But running pacrunner, it would still not be able to use authenticated proxies, so connections behind authenticated proxies would remain in _ready_ state.

> Can we find the service picked as default gateway?

Sorry, I missed this part.

> We may want to think about also preferring services that have a gateway configured (online > gateway > ready).

I think that makes sense.

knuton: > If I understand it correctly: if we don’t run pacrunner, connections behind a proxy will remain in _ready_ state. But running pacrunner, it would still not be able to use authenticated proxies, so connections behind authenticated proxies would remain in _ready_ state.

True, that would seem to be a consequence of the way connman does things.

guyonvarch: From `doc/overview-api.txt`:

> Within Connection Manager many services can be connected at the same time and also have an IP assignment. However only one can have the default route. The service with the default route will always be sorted at the top of the list.

https://git.kernel.org/pub/scm/network/connman/connman.git/tree/doc/overview-api.txt#n116

It would have been great to have an explicit flag, but it seems the right solution is to pick the first _online_ or _ready_ service.

controller/proxy/proxy.ml
guyonvarch: Picking the default service in https://github.com/dividat/playos/pull/54/commits/65109fc312bc6f134bd3b2c9046e778b4f2f4c41

kiosk/kiosk_browser/proxy.py
guyonvarch: Picking the default service in https://github.com/dividat/playos/pull/54/commits/65109fc312bc6f134bd3b2c9046e778b4f2f4c41

knuton: Ah, so you had it right all along. 😬 

Cool!

guyonvarch: I don’t remember if I did it for this reason though! It missed some documentation that is now added. :)

knuton:
### Functional Testing

- [ ] Interface
  - [ ] Add proxy
  - [ ] Update proxy
  - [ ] Remove proxy
  - [ ] Add proxy with credentials
  - [ ] Add malformed proxy
  - [ ] UI highlights active proxy
  - [ ] UI doesn't highlight inactive proxy
- [ ] Proxy functionality in kiosk
  - [ ] Initial proxy works
  - [ ] Updated proxy works
  - [ ] Removed proxy is respected
  - [ ] Authenticated proxy can be used
  - [ ] No attempt to resolve loopback/localhost through proxy
  - [ ] Captive portal detection works
- [ ] Controller
  - [ ] ZeroTier ID retrieval works when proxy configured
  - [ ] OS update works
    - [ ] with proxy
    - [ ] with authenticated proxy
    - [ ] without proxy

kiosk/kiosk_browser/proxy.py
```diff
@@ -0,0 +1,124 @@
...
+        # Update just after monitoring is on, so that we do not miss any proxy
+        # modification that could have happen before.
+        self._update(get_current_proxy(self._bus))
+
+        loop = GLib.MainLoop()
+        loop.run()
+
+    def _on_property_changed(self, *args, **kwargs):
+        if len(args) >= 2 and args[0] == 'Proxy':
+            self._update(extract_manual_proxy(args[1]))
```
knuton, line 112: I think this causes a problem because it will update the proxy configuration to that of _any_ service for which a property changes.

This means, to give a very relevant example, that when connecting a Senso to the Play Computer after it has already started, the proxy configuration will be unset in the kiosk.

Maybe the following would be better?

```suggestion
            self._update(get_current_proxy(self._bus))
```

knuton (changes requested):

knuton:
In my functional testing I found one small, though significant, defect: https://github.com/dividat/playos/pull/54/files/65109fc312bc6f134bd3b2c9046e778b4f2f4c41#diff-0c35ff9f4ea09d8857482278bdb59753efd594443d9222a36dccfb2290036c42

Should be easy to fix and then this seems ready to merge!

guyonvarch:
Wow, you did a complete test set! :+1:

> Should be easy to fix and then this seems ready to merge!

Nicely spotted! Committed in https://github.com/dividat/playos/pull/54/commits/c456fefb0bf6bb5df09833d5043c89acdfb06d87.

guyonvarch:
I checked running the kiosk that modifying the proxy of a non connected service no longer changes the current proxy.

knuton (approved):
Verified that the last addition maintains correct proxy use when an additional Ethernet connection becomes active (Senso).

Woohoo!

PR #56, opened 2020-12-10
Robust wpa_supplicant startup

knuton:
Restores wifi connectivity after upgrade to NixOS 20.09.

@guyonvarch I think you are right that this makes the dummy wifi workaround unnecessary. We should verify and suggest the connection upstream.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/bootloader/rescue/default.nix
+++ b/bootloader/rescue/default.nix
@@ -95,8 +95,6 @@ in
       # enable wpa_supplicant
       wireless = {
         enable = true;
-        # Add a dummy network to make sure that wpa_supplicant.conf is created (see https://github.com/NixOS/nixpkgs/issues/23196)
-        networks."12345-i-do-not-exist"= {};
       };
     };
 
--- a/installer/configuration.nix
+++ b/installer/configuration.nix
@@ -54,8 +54,6 @@ with lib;
     # enable wpa_supplicant
     wireless = {
       enable = true;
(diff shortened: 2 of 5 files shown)


guyonvarch:
Are you referring to the same connectivity problems mentioned in https://github.com/dividat/playos/pull/54#pullrequestreview-549082325 ?

The dummy network is also used in:

- `bootloader/rescue/default.nix`
- `installer/configuration.nix`

I’ll test it on a live system to verify the wifi connectivity.

knuton:
> Are you referring to the same connectivity problems mentioned in [#54 (review)](https://github.com/dividat/playos/pull/54#pullrequestreview-549082325) ?

Yes, I expect that those problems are in `develop` (and hopefully fixed in this PR) and have nothing to do with your work on proxies.

guyonvarch:
I just performed a test on a live USB system, I selected a WIFI network, filled the form, and got connected without problem. I could then use Play.

knuton:
> The dummy network is also used in:
> 
> * `bootloader/rescue/default.nix`
> * `installer/configuration.nix`

Thanks, I updated the PR to remove it there as well: https://github.com/dividat/playos/commit/a11ee631eed10d45deafb14e69f21d39c6fb109d

guyonvarch (approved):
Looks good to me. Is there a reason to keep it as a Draft ?

knuton:
> Is there a reason to keep it as a Draft ?

Not anymore, just added some more docs.

PR #58, opened 2021-02-03
Add status screen on tty

krksgbr:
[Trello card](https://trello.com/c/ohTYDZdL/796-status-screen-for-playos)

This adds a simple status screen to `tty8`.

## Checklist

-   [X] Changelog updated
-   ~[ ] Code documented~


system/modules/playos-status.nix
```diff
@@ -0,0 +1,54 @@
...
+  tty = "tty${toString ttyNumber}";
+  ttyPath = "/dev/${tty}";
+  driverUrl = "http://127.0.0.1:8382";
+
+  script =
+    pkgs.writeShellScriptBin "print-status" ''
+      while :; do
+        printf "\033c"
+        screen=$(xrandr --current | grep '*' | awk '{print $1}')
+        network=$(connmanctl technologies | grep 'Type\|Connected')
```
krksgbr, line 13: This prints information about all network interfaces. Should we filter for only WLAN?

system/modules/playos-status.nix
```diff
@@ -0,0 +1,54 @@
...
+let
+  ttyNumber = 8;
+  tty = "tty${toString ttyNumber}";
+  ttyPath = "/dev/${tty}";
+  driverUrl = "http://127.0.0.1:8382";
+
+  script =
+    pkgs.writeShellScriptBin "print-status" ''
+      while :; do
+        printf "\033c"
```
knuton, line 11: Maybe acquiring the info could be moved before the screen reset to remove/minimize the blank screen period?

knuton (commented):
Nice, had a quick look in a VM!

Will test on real system during course of the week.

system/modules/playos-status.nix
krksgbr: :+1:

system/modules/playos-status.nix
```diff
@@ -0,0 +1,54 @@
...
+        DISPLAY = ":0";
+      };
+      path = with pkgs; [
+        connman
+        curl
+        gnugrep
+        gawk
+        jq
+        xorg.xrandr
+      ];
```
guyonvarch, line 42: This is fine, but those could be directly used inside the script, for ex.: `${connman}/bin/connman`.

krksgbr: True! I did it like this because think the script is more readable this way.

guyonvarch: It’s not bad either with `{pkg}/bin/pkg`, because there is syntax highlighting. Now, we would not use it for `printf` for ex. So, this may be more coherent like this.

system/modules/playos-status.nix
knuton: One context where this output will be used is to quickly check practical wifi availability when a Play Computer is combined with external antennae (typically when built into the Senso's tin console).

So it would be good to verify that wifi is not only available as a technology, but that actual wifi networks can be found. I inquired about the level of detail needed, but at a minimum I think we should count and print the number of available wireless networks. (Median) Signal strength might be of interest too, but I am not sure how effortful that is to obtain and whether it is worth it.

krksgbr: Alright, I'll check what can be with reasonable effort.

krksgbr: We can easily show the number of networks found. To calculate the median signal strength, I'd rewrite the script in Python, because it becomes too cumbersome to do in bash.

Another idea I had is that we could print the information to the tty from the controller, since it already has all the info related to networking.

knuton: > Another idea I had is that we could print the information to the tty from the controller, since it already has all the info related to networking.

Interesting idea. How does that work, printing to a device file?

I guess one downside would be that getting the output from xrandr etc. would be a tad more cumbersome to do from OCaml?

guyonvarch: > Interesting idea. How does that work, printing to a device file?

I understood it differently, that it could be an HTTP request to the controller server.

krksgbr: > Interesting idea. How does that work, printing to a device file?

Users in the `tty` group can simply write to the device file.

> I guess one downside would be that getting the output from xrandr etc. would be a tad more cumbersome to do from OCaml?

I've written very little OCaml, but it seems easy enough to do it with `Lwt_process`:
https://ocsigen.org/lwt/4.1.0/api/Lwt_process

> I understood it differently, that it could be an HTTP request to the controller server.

We could do that. The controller could expose an endpoint for information relevant to the status screen. The script could call that and merge it with data it collects from `xrandr` and the rest. Though I think maybe it would be simpler to do everything in one place.

Another idea I had is to let the controller serve an html page with the status info. It could then be viewed with a text-based browser like `w3m` on the tty. This is what `services.showManual` does in nixos.
An advantage of this would be that the content could be scrollable, if we'd want to show that much info (for example, a detailed list of all wifi networks). Another advantage would be that data collection could be done on page request (users would have to manually refresh), instead of at an interval, so it would be more resource efficient (not sure how much this matters in practice).

knuton: Some quick thoughts:

- The HTML page + `w3m` approach seems attractive but [is a security problem](https://github.com/NixOS/nixpkgs/issues/83157). One would have to run a simple tool that is understood to be free of ways of executing commands or browsing the filesystem. `cat http://localhost:3333` comes to mind but isn't much different from the controller writing to the tty.
- I like the idea of writing from the controller to the tty. It does have the advantage of being able to rely on information that the controller gathers anyway.
- Anything that relies on the controller to control the content on the status tty has the downside that the tty becomes useless if the controller goes down. At the very least we would have to ensure that stale information is removed from the tty in that event.

All in all, I sense that while there may be potential to build something even more useful by relying on tools that can gather even more information, we can do something simple and already useful with the current approach, which seems dead simple and likely to be very robust.

guyonvarch:
> All in all, I sense that while there may be potential to build something even more useful by relying on tools that can gather even more information, we can do something simple and already useful with the current approach, which seems dead simple and likely to be very robust.

:+1:

One information we could add to inform the user that the printed information is fresh is _Last updated at_. One drawback I see is that it would be in UTC ? But I think it’s good nonetheless.

system/modules/playos-status.nix
krksgbr: > The HTML page + w3m approach seems attractive but is a security problem. One would have to run a simple tool that is understood to be free of ways of executing commands or browsing the filesystem. cat http://localhost:3333 comes to mind but isn't much different from the controller writing to the tty.

Nice catch!

> Anything that relies on the controller to control the content on the status tty has the downside that the tty becomes useless if the controller goes down. At the very least we would have to ensure that stale information is removed from the tty in that event.

We could launch a one-off script with `systemd` to replace the contents of the tty with a warning if the controller goes down.

> All in all, I sense that while there may be potential to build something even more useful by relying on tools that can gather even more information, we can do something simple and already useful with the current approach, which seems dead simple and likely to be very robust.

I agree! For now I'll just refine the network info and later on we can look into re-implementing the functionality as part of the controller.

krksgbr:
> One information we could add to inform the user that the printed information is fresh is _Last updated at_. One drawback I see is that it would be in UTC ? But I think it’s good nonetheless.

Good idea!

krksgbr:
> Include timezone to avoid confusion about UTC?

:+1:

knuton:
Can you add a brief section in the manual? Just a mention that a status screen is available at tty8, how to get there, how to get back.

I think either before or after the administraction chapter might be good: https://github.com/dividat/playos/tree/develop/docs/user-manual#administration

system/modules/playos-status.nix
```diff
@@ -0,0 +1,57 @@
...
+        networkConnected=$(connmanctl services | grep wifi | grep "*" | awk '{ print $2 }')
+        rfid=$(curl -s "${driverUrl}/rfid/readers" | jq -r ".readers" | jq length)
+        controller=$(systemctl is-active playos-controller)
+        time=$(date +'%T %Z')
+        printf "\033c"
+        printf "%s\n" \
+          "Screen dimensions: $screen" \
+          "Wifi networks found: $networkCount" \
+          "Connected to network: $networkConnected" \
+          "RFID: $rfid" \
```
knuton, line 22:
```suggestion
          "RFID readers found: $rfid" \
```

system/modules/playos-status.nix
```diff
@@ -0,0 +1,57 @@
...
+        screen=$(xrandr --current | grep '*' | awk '{print $1}')
+        networkCount=$(connmanctl services | grep wifi | wc -l)
+        networkConnected=$(connmanctl services | grep wifi | grep "*" | awk '{ print $2 }')
+        rfid=$(curl -s "${driverUrl}/rfid/readers" | jq -r ".readers" | jq length)
+        controller=$(systemctl is-active playos-controller)
+        time=$(date +'%T %Z')
+        printf "\033c"
+        printf "%s\n" \
+          "Screen dimensions: $screen" \
+          "Wifi networks found: $networkCount" \
```
knuton, line 20: Vanity change :wink: 

```suggestion
          "Wi-Fi networks found: $networkCount" \
```

system/modules/playos-status.nix
```diff
@@ -0,0 +1,57 @@
...
+          "Connected to network: $networkConnected" \
```
knuton, line 21: This looks a bit ambiguous when not connected. I wonder whether we can just remove this line. Typical use of this screen is only concerned with availability, not actual connection.

system/modules/playos-status.nix
```diff
@@ -0,0 +1,57 @@
...
+  ttyPath = "/dev/${tty}";
+  driverUrl = "http://127.0.0.1:8382";
+
+  script =
+    pkgs.writeShellScriptBin "print-status" ''
+      while :; do
+        screen=$(xrandr --current | grep '*' | awk '{print $1}')
+        networkCount=$(connmanctl services | grep wifi | wc -l)
+        networkConnected=$(connmanctl services | grep wifi | grep "*" | awk '{ print $2 }')
+        rfid=$(curl -s "${driverUrl}/rfid/readers" | jq -r ".readers" | jq length)
```
knuton, line 14: I realized that this will not work. The reason is that the driver only looks for readers when there is a WebSocket listener.

A simple solution might be [`pcsc_scan`](http://ludovic.rousseau.free.fr/softwares/pcsc-tools/) which is available through `pcsctools` in Nixpkgs.

The output of `pcsc_scan` looks like this:

```
❖ pcsc_scan -r
Using reader plug'n play mechanism
Scanning present readers...
0: Yubico YubiKey OTP+FIDO+CCID 00 00
1: Alcor Micro AU9560 01 00
```

knuton (changes requested):
Works well!

RFID check needs fixing, then with added docs this seems good to go. :+1:

system/modules/playos-status.nix
krksgbr: I ran into an issue with this. `pcsc_scan -r` doesn't return if there are no readers connected:
```
$ pcsc_scan -r
Using reader plug'n play mechanism
Scanning present readers...
Waiting for the first reader...
```

I found [`opensc-tool`](https://htmlpreview.github.io/?https://github.com/OpenSC/OpenSC/blob/master/doc/tools/tools.html#opensc-tool), which could be used with the `--list-readers` option. What do you think?

knuton: > I found [`opensc-tool`](https://htmlpreview.github.io/?https://github.com/OpenSC/OpenSC/blob/master/doc/tools/tools.html#opensc-tool), which could be used with the `--list-readers` option. What do you think?

I am not familiar with it, but looks fine at a quick glance.

system/modules/playos-status.nix
krksgbr: I removed it.

system/modules/playos-status.nix
krksgbr: :+1:

system/modules/playos-status.nix
krksgbr: :+1:

system/modules/playos-status.nix
krksgbr: Fixed this using `opensc-tool`.

krksgbr:
> Can you add a brief section in the manual? Just a mention that a status screen is available at tty8, how to get there, how to get back.

I've added it to the manual.

system/modules/playos-status.nix
```diff
@@ -0,0 +1,55 @@
...
+  tty = "tty${toString ttyNumber}";
+  ttyPath = "/dev/${tty}";
+  driverUrl = "http://127.0.0.1:8382";
+
+  script =
+    pkgs.writeShellScriptBin "print-status" ''
+      while :; do
+        screen=$(xrandr --current | grep '*' | awk '{print $1}')
+        networkCount=$(connmanctl services | grep wifi | wc -l)
+        rfid=$(opensc-tool --list-readers)
```
knuton, line 13: Because this output has multiple lines the indentation below will only work on the first line. A quick fix could be piping into `pr -T -o 2` (from `coreutils`).

system/modules/playos-status.nix
```diff
@@ -0,0 +1,55 @@
...
+        controller=$(systemctl is-active playos-controller)
+        time=$(date +'%T %Z')
+        printf "\033c"
+        printf "%s\n" \
```
knuton, line 17: Since we don't display a whole lot of data at the moment, `\n\n` for empty dividing lines could be nice for legibility.

knuton (changes requested):
All good from a functional perspective, suggesting some formatting changes.

system/modules/playos-status.nix
krksgbr: Good idea, thanks!

system/modules/playos-status.nix
krksgbr: :+1:

knuton (approved):
👍

PR #59, opened 2021-02-08
Improve controller's robustness against exceptions

knuton:
Makes the controller more robust against exceptions by

- encapsulating use of Unix lib to read from files,
- adding an outermost try-with as a last resort in the Opium middleware.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/bindings/util/util.ml
+++ b/controller/bindings/util/util.ml
@@ -15,6 +15,9 @@ let read_from_file log_src path =
       in
       fail exn
     | exn ->
+      let%lwt () = Logs_lwt.err ~src:log_src
+        (fun m -> m "failed to read from %s: %s" path (Printexc.to_string exn))
+      in
       fail exn
   else
     fail (Failure ("File does not exist: " ^ path))
@@ -35,4 +38,7 @@ let write_to_file log_src path str =
     in
     fail exn
   | exn ->
+    let%lwt () = Logs_lwt.err ~src:log_src
+      (fun m -> m "failed to write to %s: %s" path (Printexc.to_string exn))
+    in
     fail exn
--- a/controller/bindings/util/util.ml
+++ b/controller/bindings/util/util.ml
@@ -27,11 +27,12 @@ let write_to_file log_src path str =
     let%lwt bytes_written =
       Lwt_unix.write_string fd str 0 (String.length str)
     in
-    let%lwt () = Lwt_unix.close fd in
-    Lwt.return true
+    Lwt_unix.close fd
   with
-  | Unix.Unix_error (err, fn, _) ->
+  | (Unix.Unix_error (err, fn, _)) as exn ->
     let%lwt () = Logs_lwt.err ~src:log_src
       (fun m -> m "failed to write to %s: %s" path (Unix.error_message err))
     in
-    Lwt.return false
+    fail exn
+  | exn ->
+    fail exn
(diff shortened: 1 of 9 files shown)


controller/bindings/util/util.ml
```diff
@@ -7,15 +7,17 @@ let read_from_file log_src path =
...
     with
-    | Unix.Unix_error (err, fn, _) ->
+    | (Unix.Unix_error (err, fn, _)) as exn ->
       let%lwt () = Logs_lwt.err ~src:log_src
         (fun m -> m "failed to read from %s: %s" path (Unix.error_message err))
       in
-      Lwt.return None
+      fail exn
+    | exn ->
+      fail exn
```
guyonvarch, line 18: This exception could be hidden, in case we use `Lwt_result.catch` and `Base.Result.ok`. Why not also logging it ?

controller/server/gui.ml
```diff
@@ -64,7 +53,9 @@ let success msg =
 let error_handling =
   let open Opium_kernel.Rock in
   let filter handler req =
-    match%lwt handler req |> Lwt_result.catch with
+    (* Catch any exceptions that previously escaped Lwt *)
+    let res = try handler req with exn -> Lwt.fail exn in
```
guyonvarch, line 57: It’s fixing this manually added failure added in the info handler:

```ocaml
module InfoGui = struct
  let build app =
    app
    |> get "/info" (fun _ ->
        let _ = List.nth [] 0 in
        let%lwt server_info = Info.get () in
        page "info" [ "server_info", server_info |> Info.to_json ])
end
```

Nice! :+1:

Before the fix, It looks the same as the error you got when testing the proxy. The controller is still runnnig, but was failing in a request handler.

If I’m putting my failure code in the handler creation part, it’s making the controller to fail though:

```ocaml
module InfoGui = struct
  let build app =
    let _ = List.nth [] 0 in
    app
    |> get "/info" (fun _ ->
        let%lwt server_info = Info.get () in
        page "info" [ "server_info", server_info |> Info.to_json ])
end
```

controller/bindings/zerotier/zerotier.ml
```diff
@@ -8,15 +10,9 @@ let base_url =
...
-      "/var/lib/zerotier-one/authtoken.secret"
-  in
-  let%lwt authtoken = Lwt_io.read ic in
-  let%lwt () = Lwt_io.close ic in
-  authtoken
-  |> String.trim
-  |> return
+  Util.read_from_file
+    log_src
+    "/var/lib/zerotier-one/authtoken.secret"
```
guyonvarch, line 15: Good simplifications. :+1:

guyonvarch (commented):
Good move. :+1:

I added some remarks.

controller/server/gui.ml
knuton: > If I’m putting my failure code in the handler creation part, it’s making the controller to fail though:

Yes, that's the scary bit. If it's an intermittent failure, systemd will be able restart it.

controller/bindings/util/util.ml
knuton: 82b06a6ddc09d4aa633590ee9fbd5f0d1e63355d

controller/server/gui.ml
guyonvarch: There are also other potential points of failure, for example, an error in PlayOS.

One potential step forward, to reduce potential upgrade failures, could be to extract upgrade utility from the rest. It would complicate information on how the upgrade is going on though, from the controller.

guyonvarch (approved):

controller/server/gui.ml
knuton: Yes, we could think about that at some point.

PR #60, opened 2021-02-09
Disable autoplay policy restrictions

knuton:
The autoplay policy has now landed in QtWebEngine and we need to set the policy to allow-all to avoid delayed or missing sound in the application.

https://wiki.qt.io/QtWebEngine/ChromiumVersions

## Checklist

-   [x] Changelog updated
-   [x] Code documented


krksgbr (approved):
Tested the functionality against `develop` and sounds work as expected.
On `develop` the sound only plays after some user interaction (eg. no sound during the splash screen), while on this branch sound plays automatically.

PR #63, opened 2021-03-04
Experiment using tyxml to generate HTML

guyonvarch:
Generate the info HTML with tyxml:
https://github.com/ocsigen/tyxml

Producing HTML can be done 3 ways:

- in Ocaml, without ppx: https://github.com/ocsigen/tyxml/blob/master/examples/basic_website/site_html.ml
- in Ocaml, with ppx: https://github.com/ocsigen/tyxml/blob/master/examples/basic_website_ppx/site_html.ml
- in Reason, with JSX: https://github.com/ocsigen/tyxml/blob/master/examples/basic_website_jsx/site_html.re

I tried to use the `tyxml-jsx` at first because it is the simpler one to read, but it is only usable with `reason`. The one not using ppx is ok, but not as readable as the one using `ppx`.

What I like:

- we can compose the views directly with Ocaml,
- it’s compiled, there is no file loaded each time we change a page, and there is no potential runtime errors on the loading part,
- it makes sure images have an alt attribute (nice to have).

What is causing some trouble:

- it checks that the HTML is valid. It’s a good intention, but sometimes it can go in the way. For ex., newlines are not accepted in `<dl>`, that’s the reason I am using comments. I think it could be fixed, because it happened with `<select>`: `https://github.com/ocsigen/tyxml/issues/159`,
- types are complicated as a newcomer. I end up copying what the compiler is waiting for.

I am globally positive about it, but there are some things in the way still.


knuton:
> * it checks that the HTML is valid. It’s a good intention, but sometimes it can go in the way. For ex., newlines are not accepted in `<dl>`, that’s the reason I am using comments. I think it could be fixed, because it happened with `<select>`: `https://github.com/ocsigen/tyxml/issues/159`,

This problem wouldn't apply to the PPX-free version, right? I would actually be interested in trying that. An advantage to it is that there is no new syntax to learn, and we are already familiar with the style from Elm.

guyonvarch:
> This problem wouldn't apply to the PPX-free version, right? I would actually be interested in trying that. An advantage to it is that there is no new syntax to learn, and we are already familiar with the style from Elm.

Exact, the newline problem does not apply writing directly the tags, without the ppx preprocess. I find it a little harder to write, because there are some additional errors: for example, when one tag can have only one child (unary instead of star), there is an error that is triggered. But it may be simpler overall.

https://github.com/dividat/playos/pull/63/commits/d830f324c350c5c692682d93102dd98f9e816bf6#diff-04e11676702f7591da09313510db815fcd066add1987f5da4028c71ec46164d2

It’s still harder to write in comparison to Elm, because the produced HTML must be valid. It’s triggering compilation error than are not very good when it’s not the case. But we get used to it relatively fast.

knuton:
> It’s still harder to write in comparison to Elm, because the produced HTML must be valid. It’s triggering compilation error than are not very good when it’s not the case.

Oh wow, yes, that's very typed. :-)

I would be fine with either, probably slightly leaning towards the function-based style (but that's while not having written any yet).

One question: What kind of escape hatches are there if we wanted to produce some element or attribute not known to the lib?

guyonvarch:
> I would be fine with either, probably slightly leaning towards the function-based style (but that's while not having written any yet).

I also prefer the function-based style finaly. It is slightly harder to start with it, but there are less surprises in the end. In fact, both styles can be mixed, though we’d better stick to one style for consitency I think.

> One question: What kind of escape hatches are there if we wanted to produce some element or attribute not known to the lib?

There is `Unsafe.data` to write whatever we want. That’s what would be needed to show the changelog:

```ocaml
let generic_option current_id (id, name) =
  if current_id = Some id then
    Unsafe.data ("<option value=\"" ^ id ^ "\" selected>" ^ name ^ "</option>")
  else
    Unsafe.data ("<option value=\"" ^ id ^ "\">" ^ name ^ "</option>")
```

guyonvarch:
> In fact, both styles can be mixed, though we’d better stick to one style for consitency I think.

Well, on rendering the page in `view.ml`, it’s easier with _ppx_. If used sparingly, it may be ok.

guyonvarch:
Closing, the experiment looks good.

PR #66, opened 2021-03-15
Static IP support

krksgbr:
[Trello card](https://trello.com/c/0M7ApRSF/685-static-ip-support-in-playos)

---

This PR adds support for setting up static IP and nameservers.

For simplicity the forms to configure these are currently shown only when a service is already connected. Is it necessary to be able to configure these during the initial connection too?

The IP address inputs are validated on the client side using html5 validation with regex.

Removing the static IP configuration resets the method to `dhcp`, but does not clear the `local_address`, `netmask`, and `gateway` settings of connman (there doesn't seem to be a way of clearing these values), but they become ineffective when `method` is reset to `dhcp`.

When removing a service altogether, the IPv4 method is also reset to `dhcp` and nameservers are removed from the service.

We discussed with @knuton that we can go ahead with the [release of PlayOS](https://trello.com/c/zFs6mLJu/876-release-playos-2130) without this, if more iterations will be needed to complete this work.

## Checklist

-   [X] Changelog updated
-   [X] Code documented

### DIFF
--- a/controller/bindings/connman/connman.ml
+++ b/controller/bindings/connman/connman.ml
@@ -379,25 +379,25 @@ struct
     set_property service "Proxy.Configuration" dict
 
 
-  let set_ipv4 ?address ?netmask ?gateway service ~method'  =
-    let values =
-      List.fold_right (fun (name, value) acc ->
-          match value with
-          | Some v -> acc @ [ (name, OBus_value.C.(make_single basic_string) v) ]
-          | None -> acc
-        ) [ ("Method", Some method')
-          ; ("Address", address)
-          ; ("Netmask", netmask)
-          ; ("Gateway", gateway)
-          ]
-        []
-    in
+  let set_manual_ipv4 service ~address ~netmask ~gateway =
     let dict =
       OBus_value.C.make_single
         (OBus_value.C.(dict string variant))
-        values
-     in
-     set_property service "IPv4.Configuration" dict
+        [ ("Method", OBus_value.C.(make_single basic_string) "manual")
+        ; ("Address", OBus_value.C.(make_single basic_string) address)
+        ; ("Netmask", OBus_value.C.(make_single basic_string) netmask)
+        ; ("Gateway", OBus_value.C.(make_single basic_string) gateway)
+        ]
+    in
+    set_property service "IPv4.Configuration" dict
+
+  let set_dhcp_ipv4 service =
+    let dict =
+      OBus_value.C.make_single
+        (OBus_value.C.(dict string variant))
+        [("Method", OBus_value.C.(make_single basic_string) "dhcp")]
+    in
+    set_property service "IPv4.Configuration" dict
 
   let set_nameservers service nameservers =
     let config =
--- a/controller/bindings/connman/connman.ml
+++ b/controller/bindings/connman/connman.ml
@@ -276,7 +276,13 @@ struct
   ; strength : int option
   ; favorite : bool
   ; autoconnect : bool
+
+  (* The ipv4 field represents the actual system configuration,
+     while the ipv4_user_config represents values manually configured by the user.
+  *)
   ; ipv4 : IPv4.t option
+  ; ipv4_user_config : IPv4.t option
+
   ; ipv6 : IPv6.t option
   ; ethernet : Ethernet.t
   ; proxy : string option
@@ -322,12 +328,12 @@ struct
         _ -> None
     in
     CCOpt.(
-      pure (fun name type' state strength favorite autoconnect ipv4 ipv6 ethernet proxy nameservers ->
+      pure (fun name type' state strength favorite autoconnect ipv4 ipv4_user_config ipv6 ethernet proxy nameservers ->
           { _proxy = OBus_proxy.make (OBus_context.sender context) path
           ; _manager = manager
           ; id = path |> CCList.last 1 |> CCList.hd
           ; name ; type'; state; strength; favorite; autoconnect
-          ; ipv4; ipv6; ethernet; proxy; nameservers
+          ; ipv4; ipv4_user_config; ipv6; ethernet; proxy; nameservers
           })
       <*> (properties |> List.assoc_opt "Name" >>= string_of_obus)
       <*> (properties |> List.assoc_opt "Type" >>= string_of_obus >>= Technology.type_of_string)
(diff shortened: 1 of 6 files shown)


controller/bindings/connman/connman.mli
```diff
@@ -123,6 +125,10 @@ module Service : sig
 
   val set_manual_proxy : t -> string -> unit Lwt.t
 
+  val set_ipv4 : ?address:string -> ?netmask:string -> ?gateway:string -> t -> method':string -> unit Lwt.t
```
krksgbr, line 128: The `address`, `netmask` and `gateway` parameters are only necessary when `method` is being set to `manual`. I chose for simplicity here, but we could introduce a union type to improve type safety.

knuton:
Just from a cursory look: Allowing separate configuration of DNS and static IP has the advantage of allowing nameservers to be altered for DHCP-configured connections as well.

There are two potential disadvantages I think it would be worth to think about:

1. It could be confusing: When configuring a static IP, it is also necessary to configure DNS, but users may not be aware/forget.
2. It could facilitate abuse: When nameservers are obtained through DHCP, control is with the network. When nameservers can only be set up together with a static IP, more knowledge about the network is needed to get a functioning configuration.

So allowing for separate configuration might make normal usage a bit harder and abusive usage a bit easier. I checked "comparable" products and in the PlayStation it seems that the two can only be customized together. I am not quite sure I understood the coupling in the Xbox settings.

krksgbr:
You're right, the disadvantages outweigh the benefits.

I thought it could be useful to let our users configure nameservers separately, and I assumed that whoever would want to setup a static IP would be technically proficient enough to know that they also need to configure DNS. However as you pointed out, it would be better to optimize for easy/correct usage, since DNS configuration decoupled from static IP is not an important feature as of yet.

guyonvarch:
This is working as expected. :ok_hand: 

Some remarks on testing:

The error message is not helping. We could specify somewhere that it should be an IPv4.

https://user-images.githubusercontent.com/6768842/111322446-398a8700-8669-11eb-8624-d5c5a3ac5e82.mp4

For proxy settings, it’s better than before with the name _Proxy settings_ instead of _Advanced settings_. I propose to change the _Proxy_ label, because it’s redundant. Maybe just _URL_, and modifying also the message below: _In the form…_.



Same remark during the connection:

controller/bindings/connman/connman.ml
```diff
@@ -270,10 +276,17 @@ struct
   ; strength : int option
   ; favorite : bool
   ; autoconnect : bool
+
+  (* The ipv4 field represents the actual system configuration,
+     while the ipv4_user_config represents values manually configured by the user.
+  *)
   ; ipv4 : IPv4.t option
+  ; ipv4_user_config : IPv4.t option
```
guyonvarch, line 284: I am wondering if we could merge `ipv4_user_config` with `ipv4` with the following logic: `ipv4_user_config.or_else(ipv4)`.

### Not configured:

```ocaml
(ipv4 
  (((method' dhcp) (address 192.168.1.1) (netmask 255.255.255.0) (gateway (192.168.1.200)))))
(ipv4_user_config 
  ())
```

### Configured:

```ocaml
(ipv4
  (((method' manual) (address 192.168.1.100) (netmask 255.255.255.0) (gateway (192.168.1.200)))))
(ipv4_user_config
  (((method' manual) (address 192.168.1.100) (netmask 255.255.255.0) (gateway (192.168.1.200)))))
```

controller/bindings/connman/connman.mli
guyonvarch: While there is only 1 function, this function has to deal with options.

Suggesting to follow the proxy way for consistency:

- `set_dhcp_ipv4`
- `set_manual_ipv4`

Or to use a union type, and to do the same with `set_xxx_proxy` methods.

controller/gui/style.css
```diff
@@ -363,15 +364,19 @@
...
 .d-BackLink:hover {
     text-decoration: underline;
 }
 
 .d-BackLink:before {
-  content: "❮ ";
+    content: "❮ ";
+}
+
+.d-TabularNums {
```
guyonvarch, line 380: Is it unused ?

controller/server/gui.ml
```diff
@@ -300,9 +300,39 @@ module NetworkGui = struct
...
+    let%lwt () = Connman.Service.set_nameservers service [] in
+    Lwt.return (success (Format.sprintf "Removed static IP configuration of %s." service.name))
+
   (** Remove a service **)
   let remove ~(connman:Connman.Manager.t) req =
     let%lwt service = with_service ~connman (param req "id") in
+
+    (* Clear settings that might have been configured on the service. *)
+    let%lwt () = Connman.Service.set_nameservers service [] in
+    let%lwt () = Connman.Service.set_ipv4 service ~method':"dhcp" in
```
guyonvarch, line 334: I am wondering if we’d better move proxy removal here too, from the connect method:

```ocaml
(* Removing a proxy that would have been configured in the past *)
 let%lwt () = Connman.Service.set_direct_proxy service in
```

But in some cases, I am also wondering if it can happen that we don’t remove a connection manually, but it is unavailable, and we connect to another connection.

We could also move this clearing to the connection method.

What’s your thinking about this ?

controller/server/gui.ml
```diff
@@ -314,7 +344,8 @@ module NetworkGui = struct
     |> post "/network/:id/proxy/update" (update_proxy ~connman)
     |> post "/network/:id/proxy/remove" (remove_proxy ~connman)
     |> post "/network/:id/remove" (remove ~connman)
-
+    |> post "/network/:id/staticip/update" (update_static_ip ~connman)
+    |> post "/network/:id/staticip/remove" (remove_static_ip ~connman)
```
guyonvarch, line 348: Naming url: `static-ip` ?

guyonvarch (changes requested):

controller/bindings/connman/connman.ml
krksgbr: Good idea!

controller/gui/style.css
krksgbr: Yes, I'll remove it

controller/server/gui.ml
krksgbr: :+1:

controller/server/gui.ml
krksgbr: > But in some cases, I am also wondering if it can happen that we don’t remove a connection manually, but it is unavailable, and we connect to another connection.

I'm not completely certain what the significance of that would be. Should we have a quick call to discuss this?

krksgbr: Done in https://github.com/dividat/playos/pull/66/commits/9bcdff5420e559e163b07a6a5211a86176ec82dd.
I didn't implement checking whether the resets are necessary, because a service doesn't have the `ipv4` property unless it's already connected.

guyonvarch:
It’s currently possible to bypass the input format:

https://user-images.githubusercontent.com/6768842/111350784-c726a080-8682-11eb-801f-14951a388cfa.mp4

We may differentiate the regexp between single IP and multiple IPs.



Naming: “in the form `n.n.n.n`” to imitate the proxy message ?

krksgbr:
Separated the two cases of input validation and improved the formatting of the note on IP addresses.

guyonvarch (approved):
:ok_hand:

PR #67, opened 2021-03-15
Use variable online check timeouts

knuton:
The online check should time out quickly when first opening the network page, but should allow for longer probing when explicitly checking for connectivity.

The default timeout used for direct navigation to network page (200 ms) is just above the threshold for "instant" feedback[1], but subjectively just below the threshold for being annoyingly slow.

The timeout when clicking the refresh "button" is now longer and less likely to produce false negatives.

[1]: https://www.nngroup.com/articles/response-times-3-important-limits/

## Checklist

-   [ ] <del>Changelog updated</del>
-   [x] Code documented

### DIFF
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -207,7 +207,7 @@ module NetworkGui = struct
       with_timeout
         { duration = check_timeout
         ; on_timeout = fun () ->
-            let%lwt () = Logs_lwt.err (fun m -> m "Timeout reaching captive portal") in
+            let%lwt () = Logs_lwt.err (fun m -> m "Timeout reaching captive portal (%f s)" check_timeout) in
             return false
         }
         (fun () ->
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -196,9 +196,14 @@ module NetworkGui = struct
 
     let proxy = Proxy.from_default_service all_services in
 
+    let check_timeout =
+      try (Opium.Std.param req "timeout" |> float_of_string |> min 5.0 |> max 0.0)
+      with _ -> 0.2
+    in
+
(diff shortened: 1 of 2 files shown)


controller/server/gui.ml
```diff
@@ -196,9 +196,14 @@ module NetworkGui = struct
 
     let proxy = Proxy.from_default_service all_services in
 
+    let check_timeout =
+      try (Opium.Std.param req "timeout" |> float_of_string |> min 5.0 |> max 0.0)
+      with _ -> 0.2
+    in
```
guyonvarch, line 202: This is always returning `0.2` for me, it is not able to read the `timeout` query parameter.

`Opium.Std.param` seems to be reading from `Request.env`, which is empty:
https://github.com/rgrinberg/opium/blob/0.18.0/opium_kernel/router.ml#L38-L42

This question was asked to know of a better method to get query parameters:
`https://github.com/rgrinberg/opium/issues/32`

We currently don’t have `Request.query`, because we are on an older version of `opium`.

This code is working to get the `timeout`:

```ocaml
let check_timeout =
  let uri = Request.uri req in
  try (Uri.get_query_param uri "timeout"
    |> Option.get
    |> float_of_string
    |> min 5.0
    |> max 0.0
  ) with _ -> 0.2
in
```

guyonvarch (changes requested):

controller/server/gui.ml
knuton: Sorry about my testing lapse and thanks for the fix!

Applied in d51e8f058c71e62f104266b4257bbb5f6bb2697a, used the opportunity to avoid exceptions altogether.

guyonvarch (approved):

PR #68, opened 2021-04-08
Update driver to support senso flex

guyonvarch:
- Build driver from sources and use the latest version,
- Remove pcsclite pin on 1.8.23,
- Add play user to the dialout group.

## Tests

- I can use Senso flex on a live PlayOS

We should also check that RFID cards are still working, but I don’t have any RFID reader.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/pkgs/dividat-driver/default.nix
+++ b/pkgs/dividat-driver/default.nix
@@ -1,22 +1,35 @@
-{stdenv, fetchFromGitHub, buildGoPackage, pkgs}:
+{ stdenv, fetchFromGitHub, pkgs, buildGoModule }:
 
 let
-  sources = fetchFromGitHub {
+
+  channel = "develop";
+
+  version = "unstable";
+
+  releaseUrl = "https://dist.dividat.com/releases/driver2/";
+
+in buildGoModule rec {
+
+  pname = "dividat-driver";
+  inherit version;
+
+  src = fetchFromGitHub {
     owner = "dividat";
     repo = "driver";
-    rev = "6b1f9c925bdb4f91cb3b92393ed5425bb95cd351";
-    sha256 = "11xpgkbpl5l1n3xg770zjyg48h5zkwrfhw98ii7iw2lxpiis3gm7";
+    rev = "9146cbf2f540cd5aa9cea5828f83993c8629657b";
+    sha256 = "1lsh0lyjwdhk24zrryaqszl1k3356yzckzx32q7mbcvvkh17hs9q";
   };
-in
-buildGoPackage rec {
-  name = "dividat-driver-${version}";
-  version = "2.1.0";
 
-  src = "${sources}/src/dividat-driver";
-
-  goPackagePath = "dividat-driver";
-  goDeps = "${sources}/nix/deps.nix";
+  vendorSha256 = "1lvgp9q3g3mpmj6khbg6f1z9zgdlmwgf65rqx4d7v50a1m7g9a0m";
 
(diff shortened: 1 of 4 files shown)


knuton:
Thanks, looks good! I will only be able to test properly next week.

> * [ ]  Changelog updated <- Should we add a changelog entry ?

I say yes, let's mention it.

knuton:
The driver is missing version and channel, presumably because the Makefile wants to read them from the git repo, which isn't available here. Any ideas for fixing this?

We could make the variable assignments overridable with explicit env vars (`?=`): https://github.com/dividat/driver/blob/e1f4fc05dfe1b3cfb9e12cc14cd28f50cbd5bf40/Makefile#L20-L24

Functionality tested:

- [x] RFID reader works (pcsclite connection)
- [x] Senso connection works

guyonvarch:
> The driver is missing version and channel, presumably because the Makefile wants to read them from the git repo, which isn't available here. Any ideas for fixing this?

`buildGoPackage` and `buildGoModule` does not use the `Makefile`. We can use `buildFlagsArray` in the derivation though:

```nix
{
  buildFlagsArray = [''
    -ldflags=
      -X github.com/dividat/driver/src/dividat-driver/server.channel=${channel}
      -X github.com/dividat/driver/src/dividat-driver/server.version=${version}
      -X github.com/dividat/driver/src/dividat-driver/update.releaseUrl=${releaseUrl}
  ''];
}
```

With the changes in https://github.com/dividat/driver/pull/108, we’ll have to use `buildGoModules` instead of `buildGoPackage`.
https://nixos.org/manual/nixpkgs/stable/#sec-language-go

So, I suggest to wait for the change in `dividat/driver` before continuing there.

knuton:
> So, I suggest to wait for the change in `dividat/driver` before continuing there.

Yes, makes sense.

pkgs/dividat-driver/default.nix
```diff
@@ -1,18 +1,35 @@
...
-  src = fetchurl {
-    url = "https://dist.dividat.com/releases/driver2/${channel}/${version}/dividat-driver-linux-amd64-${version}";
-    sha256 = "0f5hd7mxrhsaxmwdyys1vk9z5rxm57y3w7qjlqj7l5v2v41g0vrm";
+{ stdenv, fetchFromGitHub, pkgs, buildGoModule }:
+
+let
+
+  channel = "develop";
+
+  version = "unstable";
```
guyonvarch, line 7: Should we create a new version on the driver ?

knuton: Maybe it's time for [`2.2.0-rc2`](https://github.com/dividat/driver/releases/tag/2.2.0-rc1). We have yet to settle on a future release model for the driver, now that its typical delivery is baked into PlayOS.

guyonvarch: I tagged the last commit of `develop` with [2.2.0-rc2](https://github.com/dividat/driver/releases/tag/2.2.0-rc2).

I am wondering if, in the future, the driver is only shipped with PlayOS, that it should be part of the PlayOS repository.

knuton: > I am wondering if, in the future, the driver is only shipped with PlayOS, that it should be part of the PlayOS repository.

I think it's still useful to have a separate project repo. We use it independently for development, and outside collaborators will continue to use it independently, too.

knuton (approved):
:+1:

PR #69, opened 2021-05-05
Introduce password-input webcomponent for the network passphrase

guyonvarch:
This hides the network passphrase by default, but its visibility can be
toggled with a SHOW/HIDE button.

There is one commit that is modifying margins, see individual commits for a simpler changeset.

## Demo

https://user-images.githubusercontent.com/6768842/117157707-1bb2e600-adbf-11eb-88de-a4e8aa9a38c5.mp4

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/server/view/network_details_page.ml
+++ b/controller/server/view/network_details_page.ml
@@ -52,6 +52,7 @@ let not_connected_form service =
           ]
           []
       ; details
+          ~a:[ a_class [ "d-Details" ] ]
           (summary [ txt "Proxy Settings" ])
           [ div
               ~a:[ a_class [ "d-Network__AdvancedSettingsTitle" ] ]
@@ -62,7 +63,7 @@ let not_connected_form service =
           ]
       ; input
           ~a:[ a_input_type `Submit
-          ; a_class [ "d-Button"; "d-Network__Button" ]
+          ; a_class [ "d-Button" ]
           ; a_value "Connect"
           ]
           ()
@@ -190,12 +191,13 @@ let connected_form service =
         ]
         [ input
             ~a:[ a_input_type `Submit
-            ; a_class [ "d-Button"; "d-Network__Button" ]
+            ; a_class [ "d-Button" ]
             ; a_value "Remove"
             ]
             ()
         ]
     ; details
+        ~a:[ a_class [ "d-Details" ] ]
         (summary [ txt "Proxy Settings" ])
         [ div
             ~a:[ a_class [ "d-Network__Form" ] ]
@@ -224,7 +226,7 @@ let connected_form service =
             ]
         ]
 
-    ; details (summary [ txt "Static IP" ])
+    ; details ~a:[ a_class [ "d-Details" ] ] (summary [ txt "Static IP" ])
       [ static_ip_form service ]
     ]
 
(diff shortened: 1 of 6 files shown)


controller/server/view/network_details_page.ml
```diff
@@ -45,14 +45,14 @@ let not_connected_form service =
               ~a:[ a_label_for passphrase_id ]
               [ txt "Passphrase" ]
           ]
-      ; input
-          ~a:[ a_input_type `Text
-          ; a_class [ "d-Input";  "d-Network__Input" ]
+      ; Unsafe.node "password-input"
```
etaque, line 48: I see unsafe, I scream :scream: 
Just for sake of clarity, why wasn't that implementable with just `input`?

etaque (changes requested):
Looks good!
Suggestion on input behaviour: when empty, there isn't anything indicating whether it's gonna render plain text or hidden, so one might paste a passphrase in it and be surprised to see it in plain. Maybe persist some text in the button when empty?

guyonvarch:
> Suggestion on input behaviour: when empty, there isn't anything indicating whether it's gonna render plain text or hidden, so one might paste a passphrase in it and be surprised to see it in plain. Maybe persist some text in the button when empty?

I think that it would not be possible to copy and paste a password in real conditions, whereas it’s possible to do it running only the controller. But still, I think it’s good to provide good defaults, so I propose to hide back the password later if the input is emptied: https://github.com/dividat/playos/pull/69/commits/084edaba4c1f754e9203cf3cfef66656ad3cf1c7.

controller/server/view/network_details_page.ml
guyonvarch: That’s what I initially did in https://github.com/guyonvarch/playos/commit/53c87c741ca07087a9bf38628574c896a7edd1e9#diff-15ac6edb2f3a04dc0bf427b2e2d64b6716f8c9af099237aeb36b347e7a300d29R49.

But using an input, you can’t add other elements in the DOM under it. So, the workaround was to create a `span` block, replacin the `input`, and adding the `input` as a child: https://github.com/guyonvarch/playos/commit/53c87c741ca07087a9bf38628574c896a7edd1e9#diff-5a9bbb8656c0233df5590430531857fbf4dd67ed38d5e4233388b8c3aa8e6faaR44-R51.

But the current solution don’t work with progressive enhancement. In case JS is disabled, it’s just a div, potentially with some style:



So, I am wondering if the hack was finally not that bad. Or another solution is to provide a wrapper component around an input.

guyonvarch: Going back to the progressive enhancement method: https://github.com/dividat/playos/pull/69/commits/9df73828cb7d6ddcc09392961270d03e48c4ddf8#.

There is the “hack” I mentioned, but the previous method also had some non trivial code to make it work, moving the attributes to the input:
https://github.com/dividat/playos/pull/69/commits/9df73828cb7d6ddcc09392961270d03e48c4ddf8#diff-5a9bbb8656c0233df5590430531857fbf4dd67ed38d5e4233388b8c3aa8e6faaL44-L48

The wrapper solution, which I did not tried, may be the saner implementation, but this would be more cumbersome to use.

guyonvarch:
https://user-images.githubusercontent.com/6768842/117782964-68c10d00-b242-11eb-82e2-4b69ea808808.mp4

etaque (approved):
> I propose to hide back the password later if the input is emptied:

Ok, that's good.

:green_circle:

PR #71, opened 2021-06-12
Move network interface list to network page

knuton:
Triggered by having to explain where to find MAC addresses, the list should be easier to find when placed within the network page.

Reserve the status page for actual dynamic status.

@guyonvarch I updated and then altered `reset.css`, as it used overly specific selectors like `ul[class]` or `ul[role]`. Probably this was done in good intention to only reset styles on styled elements (having a class), but in practice this leads to specificity issues and makes it a hassle to override the reset styles.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/server/view/network_list_page.ml
+++ b/controller/server/view/network_list_page.ml
@@ -30,9 +30,10 @@ type params =
   { proxy: string option
   ; is_internet_connected: bool
   ; services: Connman.Service.t list
+  ; interfaces: string
   }
 
-let html { proxy; is_internet_connected; services } =
+let html { proxy; is_internet_connected; services; interfaces } =
   Page.html ~current_page:Page.Network (
     div
       [ h1 ~a:[ a_class [ "d-Title" ] ] [ txt "Network" ]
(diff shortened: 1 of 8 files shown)


controller/server/view/network_list_page.ml
```diff
@@ -74,7 +75,14 @@ let html { proxy; is_internet_connected; services } =
...
               txt "No services available"
             else
-              ul ~a:[ a_class [ "d-NetworkList" ] ] (List.map service services)
+              ul
+                ~a:[ a_class [ "d-NetworkList" ]; a_role [ "list" ] ]
+                (List.map service services)
+          ]
+
+      ; section
+          [ h2 ~a:[ a_class [ "d-Subtitle" ] ] [ txt "Network Interfaces" ]
```
guyonvarch, line 84: Proposing to label it _Interfaces_, because we’re already in the context of _Network_:

```suggestion
          [ h2 ~a:[ a_class [ "d-Subtitle" ] ] [ txt "Interfaces" ]
```

guyonvarch (commented):
The previous version had `ul[class]` and `ol[class]`:
https://github.com/andy-piccalilli/modern-css-reset/blob/12e77f9a895255554e339b3de2c09b12f085dea3/dist/reset.css

guyonvarch:
See my updated comment review, for some reason, the final version I wrote did not get through.

controller/server/view/network_list_page.ml
knuton: I added the "Network" part on purpose, because to most non-technical users "Interfaces" may not mean anything. I aim to increase the likelihood that when non-technical staff is asked by their sysadmin to provide a MAC address they will find it.

I thought about using "devices" instead of "interfaces", but this is incorrect, as this is a list of interfaces. My hope was that "network interfaces", while somewhat redundant, would reassure the non-technical user that "yes yes, this is the network stuff".

guyonvarch (approved):

PR #73, opened 2021-08-16
Add note about non-GDPR compliant service

knuton:
Actively inform users that by enabling remote management the Play Computer's IP address will be sent to US-based ZeroTier.

This is necessary due to lack of a Data Processing Addendum on ZeroTier's part.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/gui/style.css
+++ b/controller/gui/style.css
@@ -183,12 +183,6 @@
     margin-top: 1rem;
 }
 
-.d-Network__Note {
-    font-size: 80%;
-    color: var(--color-hint);
-    margin-bottom: 1rem;
-}
-
 
 /* Switch */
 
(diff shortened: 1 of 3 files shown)


guyonvarch (changes requested):
Suggesting to put the text before the button, so that the flow is to read the text first, then enable remote management:



In the screenshot above, I removed the bottom margin below the text, otherwise the button is on its own.

knuton:
I updated the text to be both an explanation of what remote management is, as well as a disclaimer about the data sharing implications. That way it seems more appropriate to place it before the button.

I don't mind the whitespace between text and button, it still seems cohesive to me in the context of the page.

guyonvarch (approved):
> I updated the text to be both an explanation of what remote management is, as well as a disclaimer about the data sharing implications. That way it seems more appropriate to place it before the button.

:+1:

> I don't mind the whitespace between text and button, it still seems cohesive to me in the context of the page.

Well, I think that’s ok too.

PR #72, opened 2021-08-12
Switch to nixos-21.05

guyonvarch:
## Tests

- Run in qemu VM
- Run live in VirtualBox
- Install on VirtualBox and run
- Run live on a machine

## Remarks

- In live VMs, the time required to have the kiosk to appear can be long.
- In installed version in VirtualBox, startup of kiosk is fast but network is initially failing. Showing the errors below:

## Checklist

-   [x] Changelog updated
-   [x] Code documented

PR #75, opened 2021-08-30
Fix driver requests to pcscd (smart cards)

knuton:
We need to allow the driver (running as unprivileged user) to access smart cards via pcscd. Polkit integration is new in NixOS 21.05.

Restores RFID auth after #72.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/system/play-kiosk.nix
+++ b/system/play-kiosk.nix
@@ -107,5 +107,13 @@
   services.pcscd.enable = true;
   # Blacklist NFC modules conflicting with CCID (https://ludovicrousseau.blogspot.com/2013/11/linux-nfc-driver-conflicts-with-ccid.html)
   boot.blacklistedKernelModules = [ "pn533_usb" "pn533" "nfc" ];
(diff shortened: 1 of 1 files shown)


guyonvarch:
I don’t have any RFID reader to test the resolution, but the changes looks right.

In the `pcsc-lite` repository, they provide some information about polkit:
https://github.com/LudovicRousseau/PCSC/blob/master/doc/README.polkit

When the action id is `org.debian.pcsc-lite.access_card`, they have a test concerning the action reader. Do you think it would be beneficial to add this kind of test too ?

knuton:
> When the action id is `org.debian.pcsc-lite.access_card`, they have a test concerning the action reader. Do you think it would be beneficial to add this kind of test too ?

I don't think so, because we can not assume a fixed name, different USB devices could be used.

We also don't expect any reader we'd want to blacklist from being accessed.

guyonvarch (approved):

PR #76, opened 2021-09-24
Make journald use persistent storage

knuton:
Lets journald store logs in a persistent folder, to preserve logs across reboots.

While it can not be easily tested that the configured maximum is respected, journald should cap the size at 4 GB in the worst case.

https://www.freedesktop.org/software/systemd/man/journald.conf.html

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -9,6 +9,7 @@
 ## Changed
 
 - controller: Move network interface list to network page
+- os: Make system journal persist across reboots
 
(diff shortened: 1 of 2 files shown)


guyonvarch (approved):
It will be useful to get information in case a Play computer got restarted. :+1:

I verified that journalctl logs are persisted after a reboot on a PlayOS that I installed on VirtualBox.

PR #81, opened 2022-03-23
Clean up play-kiosk module

knuton:
Some small cleanups in the play-kiosk module.

The only change that should affect the built OS file is the removal of the unused TUI mixers for PulseAudio. Users can not access them and we have never had a need for them.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/system/play-kiosk.nix
+++ b/system/play-kiosk.nix
@@ -19,6 +19,11 @@
     group = "users";
   };
 
+  # System-wide packages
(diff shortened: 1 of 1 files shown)


krksgbr (approved):
Indeed, I don't see any changes relevant to the build other than the removal of the pulseaudio TUI. I ran a build just in case and it still works.

PR #82, opened 2022-03-23
Suppress RAUC status error during installation

knuton:
RAUC does not provide meaningful status information while an installation is in progress.

We default to a placeholder message to avoid the impression of error during a perfectly normal update.

rauc-downloading
rauc-installing
rauc1

## Testing

Can be tested by building and installing after setting main `default.nix`'s version to e.g. `2010.10.0`, installing to PC or VM, observing update process.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -4,6 +4,10 @@
 
 - os: Update nixpkgs channel to 21.11
 
+## Fixed
+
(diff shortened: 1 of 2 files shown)


krksgbr (approved):
Works as expected. Installed to a VM and saw this during updating:

image

PR #83, opened 2022-03-28
Prevent “Already connected” error when connecting to a network

guyonvarch:
Prevent submiting a form more than once. Once a form is submitted,
disable the submit input and show a loader.

This prevent geting the “Already connected” error, because (I think) 
the connection process is being worked on after the 1st submission.

## Loader

https://user-images.githubusercontent.com/6768842/160345478-c30f00c4-2b81-4d75-aed2-49fbd7820705.mp4

## Problem reproduction to check the fix

Add a delay when connecting to networks:

```patch
index e59bd4f..9950377 100644
@@ -268,6 +268,11 @@ module NetworkGui = struct
 
   (** Connect to a service *)
   let connect ~(connman:Connman.Manager.t) req =
+
+    let _ = Logs.warn (fun m -> m "\n\nTrying to connect, but sleeping before…\n") in
+    let%lwt () = Lwt_unix.sleep 2. in
+    let _ = Logs.warn (fun m -> m "\nOK\n") in
+
     let%lwt form_data = urlencoded_pairs_of_body req in
     let passphrase =
       match form_data |> List.assoc_opt "passphrase" with
```

On `develop`, when submitting connecting to a network multiple times, we get an “Already connected” error (at least when the password is wrong).

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/gui/client.js
+++ b/controller/gui/client.js
@@ -97,7 +97,6 @@ customElements.define(
       form.addEventListener('submit', function() {
         button.disabled = true
         button.style.color = 'transparent'
-        button.className += ' d-Button--Disabled'
         buttonParent.appendChild(spinnerParent)
       })
     }
--- a/controller/gui/client.js
+++ b/controller/gui/client.js
@@ -8,7 +8,7 @@ customElements.define(
 
       const input = this
 
-      const root = installParent(input, document.createElement('span'))
+      const root = wrap(input, document.createElement('span'))
       root.style = 'position: relative'
 
       const button = document.createElement('input')
@@ -75,7 +75,7 @@ customElements.define(
       const form = this
       const button = form.querySelector('input[type=submit]')
 
-      const buttonParent = installParent(button, document.createElement('span'))
+      const buttonParent = wrap(button, document.createElement('span'))
       buttonParent.style = `
         position: relative;
         height: fit-content;
@@ -109,7 +109,7 @@ customElements.define(
  * Useful to extend nodes that can not have children in web components, for
  * ex. inputs.
  */
-function installParent(node, newParent) {
+function wrap(node, newParent) {
     node.parentNode.replaceChild(newParent, node)
     newParent.appendChild(node)
     return newParent
--- a/controller/gui/client.js
+++ b/controller/gui/client.js
@@ -107,7 +107,7 @@ customElements.define(
 
 /* Place given node under a new parent node.
  *
- * Useful to extend nodes that can not have childreen in web components, for
+ * Useful to extend nodes that can not have children in web components, for
  * ex. inputs.
  */
 function installParent(node, newParent) {
--- a/controller/gui/client.js
+++ b/controller/gui/client.js
@@ -1,4 +1,4 @@
-/* Password input web component with a SHOW/HIDE toggle button
+/* Password input web component with a SHOW/HIDE toggle button.
  */
 customElements.define(
   'show-password',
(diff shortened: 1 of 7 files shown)


controller/gui/client.js
```diff
@@ -65,3 +61,57 @@ customElements.define(
...
+  },
+  { extends: 'form' }
+)
+
+/* Place given node under a new parent node.
+ *
+ * Useful to extend nodes that can not have childreen in web components, for
+ * ex. inputs.
+ */
+function installParent(node, newParent) {
```
krksgbr, line 113: What do you think of the name `wrap` instead? As I understand this function basically wraps `node` with `newParent`. It took me some effort to understand what `installParent` does; naming it `wrap` would make it more intuitive.

controller/gui/client.js
```diff
@@ -65,3 +61,57 @@ customElements.define(
...
+      `
+
+      const spinner = document.createElement('span')
+      spinner.className = 'd-Spinner'
+      spinnerParent.appendChild(spinner)
+
+      form.addEventListener('submit', function() {
+        button.disabled = true
+        button.style.color = 'transparent'
+        button.className += ' d-Button--Disabled'
```
krksgbr, line 100: This `className` is not being referenced in `style.css`.

controller/gui/client.js
```diff
@@ -65,3 +61,57 @@ customElements.define(
...
+        transform: translate(-50%, -50%);
+      `
+
+      const spinner = document.createElement('span')
+      spinner.className = 'd-Spinner'
+      spinnerParent.appendChild(spinner)
+
+      form.addEventListener('submit', function() {
+        button.disabled = true
+        button.style.color = 'transparent'
```
krksgbr, line 99: This style could be moved to the css file, since we're already styling disabled buttons there.

controller/gui/client.js
```diff
@@ -65,3 +61,57 @@ customElements.define(
...
+        buttonParent.appendChild(spinnerParent)
+      })
+    }
+  },
+  { extends: 'form' }
+)
+
+/* Place given node under a new parent node.
+ *
+ * Useful to extend nodes that can not have childreen in web components, for
```
krksgbr, line 110: Typo: childre**e**n.

krksgbr (changes requested):
Apologies for the delayed response. 

I reproduced the problem on `develop`, and your work here indeed does circumvent the issue in a user-friendly way. Good idea to show a loader in submit buttons.

I found a small number of details improve.

controller/gui/client.js
guyonvarch: https://github.com/dividat/playos/pull/83/commits/303345fbf6e3d2a933621a91875b99b0f1808766

controller/gui/client.js
guyonvarch: https://github.com/dividat/playos/pull/83/commits/43128e7a5b331874e66b266181da9c2948b6e3ff

controller/gui/client.js
guyonvarch: https://github.com/dividat/playos/pull/83/commits/43fd0021f1daf2c46e58d08fe4c1d3edbc026c32

controller/gui/client.js
guyonvarch: That’s right, it works with this usage. Now, if we think about another usage: disabling a button without showing a loader, we still would like to show the button content.

So, either we keep this manual style, or we introduce another style. I suggest to stick to update the style here, since it’s the only place, and we can think about using a style if this is used in another scenario, or for more complex rules in the future.

guyonvarch:
Thanks for you thorough review! I addressed your comments.

controller/gui/client.js
krksgbr: :+1:

krksgbr (approved):

PR #84, opened 2022-03-28
Improve network connection error message

guyonvarch:
When using an invalid passphrase, users got “Input/output error” message. That is not very informative to help them fix the problem.

Instead, use the message reported by ConnMan agent. In case the input is invalid, this is `invalid-key`.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -4,6 +4,7 @@
 
 - os: Update nixpkgs channel to 21.11
 - os: Disable virtual terminals that are not used by PlayOS
+- controller: Provide a more helpful error message when network connection is failing
 
 ## Fixed
 
--- a/controller/bindings/connman/connman.ml
+++ b/controller/bindings/connman/connman.ml
@@ -86,12 +86,6 @@ struct
 
   type t = input OBus_object.t
 
-  let report_error input service msg =
-    let%lwt () = Logs_lwt.err ~src:log_src
-        (fun m -> m "error reported to agent: %s" msg)
-    in
-    return_unit
-
   let request_input input service (fields: (string * OBus_value.V.single) list) =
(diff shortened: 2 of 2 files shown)


knuton:
Very good

Have you had a look at [the other error codes](https://git.kernel.org/pub/scm/network/connman/connman.git/tree/include/service.h#n73) yet?

Having a quick look:

- `CONNMAN_SERVICE_ERROR_CONNECT_FAILED`, `CONNMAN_SERVICE_ERROR_LOGIN_FAILED`, `CONNMAN_SERVICE_ERROR_AUTH_FAILED` all seem to deal with VPN provider issues
- `CONNMAN_SERVICE_ERROR_OUT_OF_RANGE`, `CONNMAN_SERVICE_ERROR_PIN_MISSING`, `CONNMAN_SERVICE_ERROR_DHCP_FAILED` I can not find to be ever returned by grepping the sources
- `CONNMAN_SERVICE_ERROR_BLOCKED` is related to a wifi condition I don't understand at first glance

So maybe this patch already takes care of the main failure code we would expect to occur our case.

guyonvarch:
> Have you had a look at the other error codes yet?

I didn’t yet. With the proposed modifications, we’re giving to the user the error reported by the agent, so hopefully, if there is another problem that we don’t reproduce yet, at least the user get a more helpful key, like `auth-failed`. Though I currently don’t know the difference between the different connection/auth errors.

My plan is to test with this modification and the other PR fixing “Already Connected” connecting to a Wifi point, in the same conditions I had “Already Connected” and “Input/Output” errors, and check if the errors are not present / provide good feedback.

knuton (approved):
Nice

PR #85, opened 2022-04-04
Fix webview dialog memory leak

guyonvarch:
- Fix webview dialog memory leak by assigning a parent to the webview in 17bddba
- Reduce memory usage by instantiating dialog webview only once in 8ba01db
- Reset main widget view to prevent activity like playing sounds in c0ec084
- Improve cleaning up after dialog is close in 9c2ffd8

## Testing

I open `htop` and look at global memory consumption. This in not very precise, because memory could grow because of another program running on my machine. Then I open and close the dialog modal repeatedly.

## Checklist

-   ~~[ ] Changelog updated~~
-   [x] Code documented

### DIFF
--- a/kiosk/kiosk_browser/main_widget.py
+++ b/kiosk/kiosk_browser/main_widget.py
@@ -1,8 +1,6 @@
 from PyQt5 import QtWidgets, QtCore
 
-from kiosk_browser import browser_widget, captive_portal
-from kiosk_browser import proxy as proxy_module
-from kiosk_browser import webview_dialog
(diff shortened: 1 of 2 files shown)


guyonvarch:
When testing a lab-key on a machine, I noticed 2 things:

- when connection is not established, closing System Settings let the overlay visible,
- and opening System Settings when it has already has been open before, we sometimes see briefly the previous view, then a blank page, then the new view.

I am working on it, sorry for putting this too fast in review.

guyonvarch:
Closing in favor of #87.

PR #86, opened 2022-04-04
Limit max resolution

krksgbr:
This xorg config seems to work to cap the resolution that X will use.
`Identifier` needs to be the name of the output, as returned by `xrandr -q`.
We'll probably need separate entries for each port on PlayOS machines (`DP, HDMI` etc).
The current config is only for testing inside a VM.

### Considerations
- The identifiers for the monitors are generated by the video card driver. I wonder if it's safe to rely on these labels, I am not sure how stable they are.

- I wonder what will happen if a TV is connected with an aspect ratio other than what we specify in `PreferredMode`. Most TVs will have a 16:9 aspect ratio, but with at home installations it's possible people might use a non-standard aspect ratio such as 21:9 (ultrawide monitors).

## Checklist

-   [X] Changelog updated
-   [X] Code documented
-   ~[ ] User manual updated~

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -15,6 +15,7 @@
 - kiosk: Show a loader when connecting to Play
 - kiosk: Show an informative error page when connecting to Play has failed
 - kiosk: Open System Settings (Ctrl+Shift+F12) and Network Login (behind captive portal) in a dialog
+- os: Limit resolution to 1080p on UHD screens
 
 ## Fixed
 
--- a/system/play-kiosk.nix
+++ b/system/play-kiosk.nix
@@ -28,6 +28,31 @@
   services.xserver = let sessionName = "kiosk-browser"; in {
     enable = true;
 
+    # Limit resolution to full HD on UHD screens.
+    # This is to avoid performance issues with games
+    # and scaling issues of the controller UI.
+    extraConfig = ''
+      Section "Monitor"
+              Identifier      "HDMI-1"
+              Option "PreferredMode"  "1920x1080"
+      EndSection
+
+      Section "Monitor"
+              Identifier      "HDMI-2"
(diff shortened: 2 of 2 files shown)


krksgbr:
I tested this config on multiple 4k screens and also a WQHD (3440x1440) screen.

**UDH screens**
The config effectively limits the resolution to 1080p on 4k screens. This was tested by reading `screen.width` and `screen.height` in the kiosk, showing `1920x1080`.

**WQHD screens**
Unfortunately this config forces `1920x1080` on WQHD screens, stretching the output:
IMG_20220422_165321

**<1080p screens**
I couldn't find a screen smaller than 1080p, so I tested in QEMU and virtualbox VMs, since their virtual displays are smaller than 1080p by default. This did not work consistently between the two VMs. In QEMU the resolution was `1920x1080p` while in virtualbox it stayed at the default `1024x768`. We can't make a good judgment based on this and it would be better to test on a real screen. Based on how the config behaves on WQHD, my prediction is that it will force `1080p` on smaller screens too.

**3D game performance**
I played all the 3d games and they ran smoothly on 4k screens.

**Layout**
Layout stayed intact and was scaled properly in the controller and Play.


**Summary**
The config achieves our main goal of limiting resolution on UHD screens but does not work correctly when it comes to non-standard screens. I am not sure if we can release this as is, or if we should look for a solution that works for all screens. I wonder if we should assume some people will use screens with non-standard aspect ratios.

An alternative solution would be to to run a script setting the resolution with `xrandr`, and/or when a screen is attached. This would allow us to choose a proper resolution for any screen, but might be more error-prone.

knuton:
While a good proof of concept, the known limitations make this unsafe to deploy widely at this time. We should look into a solution with opt-in that can be tried on individual sites initially.

PR #87, opened 2022-04-05
Fix webview dialog memory leak

guyonvarch:
This makes fewer modifications compared to #85. I didn’t know how to reset the web view of the dialog, so that when we open the dialog again, we don’t briefly see what was loaded before.

Memory usage grows faster here, but it’s getting down from time to time when repeatedly opening and closing it.

## Checklist

-   ~~[ ] Changelog updated~~
-   [x] Code documented

### DIFF
--- a/kiosk/kiosk_browser/main_widget.py
+++ b/kiosk/kiosk_browser/main_widget.py
@@ -42,15 +42,17 @@ class MainWidget(QtWidgets.QWidget):
             self._layout.addWidget(self._captive_portal_message)
 
     def _show_settings(self):
+        self._hide_browser_widget()
         webview_dialog.widget(
                 self, 
                 "System Settings", 
                 self._settings_url, 
                 additional_close_keys = [self._toggle_settings_key],
-                on_dialog_close = lambda: self._browser_widget.reload()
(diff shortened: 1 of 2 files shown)


knuton:
I liked that the application stayed in the background before https://github.com/dividat/playos/pull/87/commits/95fe137bf5e42d612634f1dac61c45e949a205cf.

It helps preserve context and may even give direct feedback that a configuration change has had an effect. Can you explain the benefits of removing it?

guyonvarch:
I also liked keeping the context below.

Here are my reasons for this change:

- On background activity: If the dialog is open while someone is connected, after some time, there would be a reminder asking if the person is still there, with a sound played. This could also be during a game. This may seems strange to open a dialog during a game, but who knows. I may forget about other activity that could occur in the background.
- Since closing the dialog is reloading the browser widget, I find it more fair to do it from the beginning. I think this is not a dialog as we’re used to see. Additionally, there is a lot coming up on closing the dialog: we see 1. the dialog, 2. the old page very briefly, and 3. the loader and the start page of Play.

knuton (approved):

PR #88, opened 2022-04-05
Multi-input proxy form for handling special characters

knuton:
This improves support for web proxies by using separate imports for server and credentials, so that users do not need to percent-encode special characters.

The core change to achieve separation and proper encoding and decoding is in the first three commits. The rest of the changeset consists of yak shaving with the aim of re-harmonizing the UI after the proxy form alterations.

The changeset mostly simplifies the handling of the configuration in the server, except for bba08bc672940f08ab4890c0f021a109e80367a5 which adds handling for the special case of proxy configuration being resubmitted without a password. Because we must not expose the password but want to combine all configuration into one form, we must allow form resubmission without a password. A check has been added to determine whether to keep the previous proxy configuration or update with altered data.

So while some of the changes in this PR are only tangentially related to the improved proxy form, they restore the UI to a cohesive whole, which is why I am opening one PR only.

An unrelated UI change has been extracted to #89.

I would like to get feedback on the changed form before updating the manual.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [ ] User manual updated


controller/server/gui.ml
```diff
@@ -249,22 +242,84 @@ module NetworkGui = struct
...
+    | None ->
+        let open Cohttp in
+        let open Cohttp_lwt_unix in
+        let%lwt () = Logs_lwt.err ~src:log_src
+          (fun m -> m "disabling static ip %s" (get_prop "static_ip_address"))
+        in
+        let%lwt () = Connman.Service.set_dhcp_ipv4 service in
+        let%lwt () = Connman.Service.set_nameservers service [] in
+        return ()
+    | Some _ ->
```
guyonvarch, line 314: Use `Option.is_some` ?

controller/server/gui.ml
```diff
@@ -249,22 +242,84 @@ module NetworkGui = struct
...
+      in
+      let password_input =
+        form_data |> List.assoc "proxy_password" |> List.hd |> non_empty
+      in
+      match host_input, port_input, user_input, password_input with
+      (* Complete configuration was submitted *)
+      | Some host, Some port, Some user, Some password ->
+        return (Some (Service.Proxy.make ~user:user ~password:password host port))
+      (* Configuration was submitted without password, check whether it is
+         safe to refill from stored value *)
```
guyonvarch, line 275: Could it lead to trouble in case someone wrongly enters a password and validate, then remember that the password must be empty for this proxy ?

Without this, It gets troublesome to update Ipv4 if proxy has been configured in the past.

Did you consider putting the password directly ? It would be still accessible, but at least not visible by default.

controller/server/view/network_details_page.ml
```diff
@@ -252,6 +218,21 @@ let html service =
...
+    form
+      ~a:[ a_action ("/network/" ^ service.id ^ "/remove")
+      ; a_method `Post
+      ; a_class [ "d-Network__Forget" ]
+      ; Unsafe.string_attrib "is" "disable-after-submit"
+      ]
+      [ input
+          ~a:[ a_input_type `Submit
+          ; a_class [ "d-Button" ]
+          ; a_value "Forget"
```
guyonvarch, line 231: I would be more comfortable with `Disconnect`. Does iOS use this naming ?

controller/proxy/proxy.ml
```diff
@@ -1,57 +0,0 @@
...
-          | Some user, Some password -> Some { user; password }
-          | _ -> None)
-        ; host
-        ; port
-        }
-    | _ -> None
-  else
-    None
-
-let to_string ~hide_password t =
```
guyonvarch, line 32: The rationale behind having to specify `~hide_password` was to enforce the caller to say if he wanted to have the password visible explicitly. Now, I am wondering if `to_uri >> to_string` could be use by inadvertance instead of `pp`.

guyonvarch (changes requested):
Working well!

I liked knowing that there was a password, previously showed with stars. Is there a specific reason not to do this anymore ?

image

Form arrangement: show password beside username ?

image

I still have to verify that I can use special characters as credentials.

controller/server/gui.ml
```diff
@@ -249,22 +242,84 @@ module NetworkGui = struct
...
-      | None -> fail_with (Format.sprintf "'%s' is not a valid proxy. It should be in the form 'http://host:port' or 'http://user:password@host:port'." proxy_str)
+  let make_proxy current_proxy_opt form_data =
+    let open Service.Proxy in
+    let non_empty s = if s = "" then None else Some s in
+    let opt_int_of_string s = try Some (int_of_string s) with _ -> None in
+
+    match form_data |> List.assoc_opt "proxy_enabled" with
+    | None ->
+        return None
+    | Some _ ->
```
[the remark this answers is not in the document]
knuton: Gave it a try, but I am not a fan. Using pattern matching allows to front-load the `None` case and each of the blocks is directly labeled with the case it treats.

Using an ifte expression seems more prone to cause some close relative of [Boolean blindness](https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/):

```diff
diff --git a/controller/server/gui.ml b/controller/server/gui.ml
index b997164..7623ad0 100644
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -246,10 +246,7 @@ module NetworkGui = struct
     let non_empty s = if s = "" then None else Some s in
     let opt_int_of_string s = try Some (int_of_string s) with _ -> None in
 
-    match form_data |> List.assoc_opt "proxy_enabled" with
-    | None ->
-        return None
-    | Some _ ->
+    if form_data |> List.assoc_opt "proxy_enabled" |> Option.is_some then
       let host_input =
         form_data |> List.assoc "proxy_host" |> List.hd |> non_empty
       in
@@ -288,6 +285,8 @@ module NetworkGui = struct
       (* Incomplete server information *)
       | _ ->
         fail_with "A host and port are required to configure a proxy server"
+    else
+      return None
 
   (** Set static IP configuration on a service *)
   let update_static_ip ~(connman: Connman.Manager.t) service form_data =
@@ -296,24 +295,23 @@ module NetworkGui = struct
           |> List.assoc s
           |> List.hd
         in
-    match form_data |> List.assoc_opt "static_ip_enabled" with
-    | None ->
-        let open Cohttp in
-        let open Cohttp_lwt_unix in
-        let%lwt () = Logs_lwt.err ~src:log_src
-          (fun m -> m "disabling static ip %s" (get_prop "static_ip_address"))
-        in
-        let%lwt () = Connman.Service.set_dhcp_ipv4 service in
-        let%lwt () = Connman.Service.set_nameservers service [] in
-        return ()
-    | Some _ ->
-        let address = get_prop "static_ip_address" in
-        let netmask = get_prop "static_ip_netmask" in
-        let gateway = get_prop "static_ip_gateway" in
-        let nameservers = get_prop "static_ip_nameservers" |> String.split_on_char ',' |> List.map (String.trim) in
-        let%lwt () = Connman.Service.set_manual_ipv4 service ~address ~netmask ~gateway in
-        let%lwt () = Connman.Service.set_nameservers service nameservers in
-        return ()
+    if form_data |> List.assoc_opt "static_ip_enabled" |> Option.is_some then
+      let address = get_prop "static_ip_address" in
+      let netmask = get_prop "static_ip_netmask" in
+      let gateway = get_prop "static_ip_gateway" in
+      let nameservers = get_prop "static_ip_nameservers" |> String.split_on_char ',' |> List.map (String.trim) in
+      let%lwt () = Connman.Service.set_manual_ipv4 service ~address ~netmask ~gateway in
+      let%lwt () = Connman.Service.set_nameservers service nameservers in
+      return ()
+    else
+      let open Cohttp in
+      let open Cohttp_lwt_unix in
+      let%lwt () = Logs_lwt.err ~src:log_src
+        (fun m -> m "disabling static ip %s" (get_prop "static_ip_address"))
+      in
+      let%lwt () = Connman.Service.set_dhcp_ipv4 service in
+      let%lwt () = Connman.Service.set_nameservers service [] in
+      return ()
 
 
   (** Connect to a service *)

```

guyonvarch:
Working well with a password containing special characters like ` :"@`.

In the VM, when using the wrong credentials, the controller is not connected anymore, but the kiosk is still connected. I wonder if it has something to do with the VM or not, and if this already existed before this changeset.

controller/server/gui.ml
knuton: > Did you consider putting the password directly ? It would be still accessible, but at least not visible by default.

What do you mean by this? I didn't want to emit it as the form value, if that's what you mean. I don't have a concrete exploitability in mind, but it just seemed to be better to avoid it.

controller/server/gui.ml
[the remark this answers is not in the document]
guyonvarch: OK. I suggested this because it’s a warning in Rust, but here we have readability benefits.

controller/server/view/network_details_page.ml
knuton: It seems to be widely used:

- iOS
- Android
- [Connman GTK](https://github.com/jgke/connman-gtk/blob/b72c6ab3bb19c07325c8e659902b046daa23c506/po/connman-gtk.pot#L597-L599)
- [Windows](https://www.lifewire.com/forget-network-in-windows-11-5208278)

I think "Forget" is better than "Disconnect" because it implies that this isn't just a temporary disconnect and all settings will be forgotten.

controller/server/gui.ml
guyonvarch: Yes, I meant to provide the password as input. It would simplify the reconnexion logic, potentially preventing being locked with the password scenario I provided. The password would still be hidden from sight without revealing it manually.

controller/server/view/network_details_page.ml
guyonvarch: OK.

controller/proxy/proxy.ml
knuton: Yeah, that seems like a fair concern.

What do you think about e6014fb?

controller/bindings/connman/connman.mli
```diff
@@ -160,4 +199,8 @@ module Manager : sig
       This uses the ConnMan ServiceChanged Signal but returns complete services instead of only changes.*)
   val get_services_signal : t -> Service.t list Lwt_react.S.t Lwt.t
 
+  (** [from_default_service sevrices] returns the [proxy] of the ready or
```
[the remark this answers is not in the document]
knuton: d833706

knuton:
> In the VM, after being connected to the proxy, then using the wrong credentials, the controller is not connected anymore, but the kiosk is still connected. I wonder if it has something to do with the VM or not, and if this already existed before this changeset.

I can observe the same behaviour in cbe237f95df1ddfb899c1d6b9824dc619d0a6ef0. Might just be the behaviour of QtWebEngine.

guyonvarch:
I pushed a small fix concerning equality: https://github.com/dividat/playos/pull/88/commits/0bfe7986c3e18197cbbb31c19b525b19276f81c6.

I’ll merge this PR, and I’ll create 2 new ones:

- add a hint when password is set in proxy,
- try to add UI elements to ease in password management of proxy forms.

PR #89, opened 2022-04-05
Mark network services that are connected

knuton:
We currently display the Internet connectivity in the overview, but do not show which network services are connected.

By adding a marker for service connectivity it should become easier to understand current configuration, especially in failure scenarios.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


guyonvarch:
Good idea.

We could even be more explicit in the UI, for example (https://github.com/guyonvarch/playos/commit/4b9bbaccd5eef5d113f23e6cb21ada0c69cdf52d):

image

knuton:
The problem I see with being more explicit is that we need to figure out what exactly to claim.

My worry is that "Connected" will be interpreted by many as "YAY INTERNET!", which is not at all what it means.

Some consumer OS show explicit information to avoid this confusion, for example iOS ([Source](https://tunecomp.net/wifi-connected-but-no-internet-access-android-windows/)):

image

I guess we could also follow this general pattern and separate connected from available networks into two different lists, without making additional explicit claims about the "connectedness" of the networks.

guyonvarch:
From connman documentation:

> The "ready" state signals a successfully connected device. "online" signals that an Internet connection is available and has been verified.

We have a handmade mechanism to detect internet connection, but we could drop it and insted, augment the information attached to a network, for example with 2 tags:

- ready: `Connected` (black) `No internet connection` (orange)
- online: `Connected` (black) `Online` (green)

knuton:
> * ready: `Connected` (black) `No internet connection` (orange)

Yes, but then we would say the Senso is connected but does not provide an Internet connection. Which is correct but probably confusing. More importantly, if a service is in `ready` state it does _not_ mean that there is no Internet connection. It merely means that connman's check did not confirm that it has an Internet connection (e.g. because the test host is not whitelisted in firewall).

> We have a handmade mechanism to detect internet connection, but we could drop it and insted

While connman's check is useful, ours is more meaningful in the end, because it contacts a host that we control and ask customers to whitelist.

I don't think that "Internet connection" is a well-defined concept, it does not mean anything specific. I imagine Apple in iOS implemented some heuristic check(s) to determine whether a network provides access in a way that users will perceive as "I have Internet".

guyonvarch:
> More importantly, if a service is in ready state it does not mean that there is no Internet connection. It merely means that connman's check did not confirm that it has an Internet connection (e.g. because the test host is not whitelisted in firewall).

We can have some problem with our manual check: sometimes, I connect to a network, but then it says there is no internet connection, potentially because of the timeout.

> While connman's check is useful, ours is more meaningful in the end, because it contacts a host that we control and ask customers to whitelist.

I think we should not discard connman for connectivity too fast. We could patch connman for example to use our own URL instead of the default one. Another benefit is that we stop making requests to the default URL specified in connman.

One advantage of connman checks over ours is that it’s not synchronous. Our check uses a short timeout, which can flag a working connection as not connected, because it’s not fast enough.

knuton:
> I think we should not discard connman for connectivity too fast. We could patch connman for example to use our own URL instead of the default one. Another benefit is that we stop making requests to the default URL specified in connman.

I have researched connman's online check to some extent before, I don't think it is a useful replacement for our own connectivity check, even if we would patch the URL (which may be useful, if only to reduce the whitelisting burden on firewall maintainers).

connman's online check runs once when a service initially becomes available and is not re-run if the network service state doesn't change (https://lists.ofono.org/hyperkitty/list/connman@lists.01.org/thread/C4BFH7RACCV3354UW64I5I7U4YPLGILT/; https://trello.com/c/ZUZoKpqp/743-playos-periodic-gateway-re-selection). It is meant specifically to help select the best default gateway, not as a signal of whether or not the device is currently online.

knuton:
Replaced dot marker by a checkmark and separated network service lists into connected and available lists.

knuton:
I reverted the updates to the manual. This needs to be redone when the UI has been settled.

Maybe it even makes sense to wait for #88 to be complete and then revise the manual in one take? I am adding a checklist to the Trello card.

knuton:
> I thought about hiding the “Available Networks” section if the list is empty and there is at least one connected network. But in case the connected network is the Senso, we’d still like to connect to another network and be explicitly told that there is no available network, right ?

Yeah, I think it's safer to be explicit here all in all.

PR #90, opened 2022-04-11
Show an indication when password is used in proxy

guyonvarch:
Context: https://github.com/dividat/playos/pull/88#issuecomment-1091902344

## Before

Without user info:

image

With user no matter if password is set:

image

## After

Without user info:

image

With user and without password:

image

With user and password:

image

## Checklist

-   ~~[ ] Changelog updated~~
-   ~~[ ] Code documented~~
-   ~~[ ] User manual updated~~ (We’ll check it before the next release)

### DIFF
--- a/controller/bindings/connman/connman.ml
+++ b/controller/bindings/connman/connman.ml
@@ -296,28 +296,19 @@ struct
       else
         None
 
-    let to_uri ~include_password t =
+    let to_uri ~include_userinfo t =
       let escape_userinfo = Uri.pct_encode ~component:`Userinfo in
       let
         userinfo =
           Option.map
-            (fun credentials ->
-              if include_password then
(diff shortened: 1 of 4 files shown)


krksgbr (approved):
I noticed a slight change in the behavior of `to_uri`, but it doesn't have any practical implications; the new behavior is actually more logical.

**Details**

Before:
```
A
~include_password: false
http://user@myhost.com:8080

B
~include_password: true
http://user:secret@myhost.com:8080
```

After:
```
A
~include_userinfo: false
http://myhost.com:8080

B
~include_userinfo: true
http://user:secret@myhost.com:8080
```

A's are different, but calling `to_uri` with `~include_password: false` was only used for showing the URL in the UI (via `Service.Proxy.pp`), and it isn't useful beyond that purpose anyway.

PR #91, opened 2022-04-13
Ease keeping previously defined proxy password

guyonvarch:
With the previous mechanism, one scenario was not possible: delete a password while keeping the same configuration otherwise. Moreover, I find it to be working out too implicitly.

This introduces a “Keep previously defined password” checkbox if a password has already been defined.

If it is checked, this means we keep the previously defined password. Otherwise, a password input is showed and we can type a new password, or none if we want to reset it.

## Screenshots

image

image

## Checklist

-   ~~[ ] Changelog updated~~
-   [x] Code documented
-   ~~[ ] User manual updated~~ (Will be done before releasing)

### DIFF
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -262,19 +262,23 @@ module NetworkGui = struct
       in
       let password = 
         match (keep_password, current_proxy_opt) with
-        | (true, Some ({ credentials = Some { password } })) -> Some password
-        | _ -> password_input
+        | (true, Some ({ credentials = Some { password } })) -> Ok (Some password)
+        | (true, _) -> Error "Failure to retrieve proxy password. Please re-submit the form."
+        | _ -> Ok password_input
       in
       match host_input, port_input, user_input, password with
       (* Configuration without credentials was submitted *)
-      | Some host, Some port, None, None ->
+      | Some host, Some port, None, Ok None ->
         return (Some (Service.Proxy.make host port))
       (* Configuration with credentials was submitted *)
-      | Some host, Some port, Some user, password ->
+      | Some host, Some port, Some user, Ok password ->
         return (Some (Service.Proxy.make ~user:user ~password:(Option.value ~default:"" password) host port))
       (* Configuration without user but with password was submitted *)
-      | _, _, None, Some _ ->
+      | _, _, None, Ok (Some _) ->
         fail_with "A user is required if a password is provided"
+      (* Password retrieval error *)
+      | _, _, _, Error msg ->
+        fail_with msg
       (* Incomplete server information *)
       | _ ->
         fail_with "A host and port are required to configure a proxy server"
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -337,8 +337,6 @@ module NetworkGui = struct
       | Some proxy -> Connman.Service.set_manual_proxy service proxy
     in
 
-    (* Grant time for changes to take effect and return to overview *)
-    let%lwt () = Lwt_unix.sleep 0.5 in
     redirect' (Uri.of_string "/network")
 
   (** Remove a service **)
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -257,13 +257,24 @@ module NetworkGui = struct
       let password_input =
         form_data |> List.assoc "proxy_password" |> List.hd |> non_empty
       in
-      match host_input, port_input, user_input, password_input with
+      let keep_password = 
+        form_data |> List.assoc_opt "keep_password" |> Option.is_some
+      in
+      let password = 
+        match (keep_password, current_proxy_opt) with
+        | (true, Some ({ credentials = Some { password } })) -> Some password
+        | _ -> password_input
+      in
+      match host_input, port_input, user_input, password with
+      (* Configuration without credentials was submitted *)
+      | Some host, Some port, None, None ->
+        return (Some (Service.Proxy.make host port))
       (* Configuration with credentials was submitted *)
       | Some host, Some port, Some user, password ->
         return (Some (Service.Proxy.make ~user:user ~password:(Option.value ~default:"" password) host port))
-      (* Configuration without credentials was submitted *)
-      | Some host, Some port, None, _ ->
-        return (Some (Service.Proxy.make host port))
+      (* Configuration without user but with password was submitted *)
+      | _, _, None, Some _ ->
+        fail_with "A user is required if a password is provided"
       (* Incomplete server information *)
       | _ ->
         fail_with "A host and port are required to configure a proxy server"
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -258,25 +258,9 @@ module NetworkGui = struct
         form_data |> List.assoc "proxy_password" |> List.hd |> non_empty
       in
       match host_input, port_input, user_input, password_input with
-      (* Complete configuration was submitted *)
-      | Some host, Some port, Some user, Some password ->
-        return (Some (Service.Proxy.make ~user:user ~password:password host port))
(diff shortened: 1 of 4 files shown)


guyonvarch:
I have a commit that I will submit after, in order to reduce the size of texts a bit:

image

krksgbr (changes requested):
This doesn't seem to work correctly for me.
The password is not kept, even if I leave the checkbox checked.

Steps to reproduce:
1. Fill out and submit the proxy form
2. Submit the proxy form again without any changes
3. Visit the page of the network again, the password field is empty. Also if I check the proxy settings using connman-gtk, the password is missing.

I tested this outside a vm, just by running the controller independently. Not sure if this can make a difference.

guyonvarch:
I don’t reproduce the problem. Using `chromium` or `firefox`, when I click `update` on this:

image

The password is not erased afterward:

image

Which browser are you using ?

krksgbr:
I used chromium.
I will try Firefox and will also do some tests in the VM.

krksgbr:
I tested in a VM and it works as expected.
I also tested again outside of a VM and could not reproduce the issue anymore. I am on a different network now, but it would be strange if that should make a difference.

krksgbr:
Update: I tested again outside a VM after making connman forget the network and the issue re-appeared. I will do another test on an actual PlayOS machine, as I can't test wifi in a VM.

guyonvarch:
> Update: I tested again outside a VM after making connman forget the network and the issue re-appeared. I will do another test on an actual PlayOS machine, as I can't test wifi in a VM.

This should also work on our machines, but I cannot reproduce the problem using the following steps:

1. Forget the network
2. Connect to the network
3. Fill out and submit the proxy form
4. Submit the proxy form again without any changes
5. Visit the page of the network again, the password field is empty. Also if I check the proxy settings using connman-gtk, the password is missing.

Am I missing a step, or the problem could maybe come from something else ?

guyonvarch:
I’ll check on a live system using a lab key.

krksgbr:
> Am I missing a step, or the problem could maybe come from something else ?

No, this is what I did to make the problem surface again. However when I did it again, it didn't produce the issue. :man_shrugging:

krksgbr:
It does work as expected most of the time, but some of the time it erases the password.
I seem to be able to reproduce it reliably by repeatedly opening the network page and just hitting the update button. When doing this a number of times, about 1 out of 5 times the password is erased. Could you also try this and see if you can observe the issue as well? Otherwise we could pair and debug together.

krksgbr:
You were right about the source of the problem.
With this diff
```diff
diff --git a/controller/server/gui.ml b/controller/server/gui.ml
index 95efdc6..6c3ba7e 100644
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -249,6 +249,8 @@ module NetworkGui = struct
     let%lwt service = with_service connman service_id in
     Lwt.return (page (Network_details_page.html service))
 
+
+  exception Weird of string
   (** Validate a proxy, fail if the proxy is given but invalid *)
   let make_proxy current_proxy_opt form_data =
     let open Service.Proxy in
@@ -277,6 +279,7 @@ module NetworkGui = struct
       let password = 
         match (keep_password, current_proxy_opt) with
         | (true, Some ({ credentials = Some { password } })) -> Some password
+        | (true, None) -> raise (Weird "This should not happen.")
         | _ -> password_input
       in
       match host_input, port_input, user_input, password with
```
and after a few tries I get:

image

guyonvarch:
Now it will fails when proxy should be retrieved but can’t: https://github.com/dividat/playos/pull/91/commits/5d6ca81107df6f900ca8449bea99ff4b9a3426a0.

I can’t reproduce the problem on my machinethough . This may be because I am only connected to one network.

Could you add the following debugging in `get_default_proxy`, just after getting the services, and look at what is happening in case the password is dropped ?

```ml
let%lwt () = Logs_lwt.warn ~src:log_src (fun m -> m "\n\n\nServices\n--------\n") in

let%lwt () = 
  services
    |> Lwt_list.iter_s (fun s ->
        let props = Sexplib.Sexp.to_string_hum (sexp_of_t s) in
        Logs_lwt.warn ~src:log_src (fun m -> m "\n%s\n" props))
in

let%lwt () = Logs_lwt.warn ~src:log_src (fun m -> m "\n\nEnd\n---\n") in
```

krksgbr:
The service I'm connected to, when things go normally:
```
((_proxy <opaque>) (_manager <opaque>)
 (id wifi_dcfb48343523_497474414d6f6b6b486f6c414d6f6b6b_managed_psk)
 (name IttAMokkHolAMokk) (type' Wifi) (state Online) (strength (90))
 (favorite true) (autoconnect true)
 (ipv4
  (((method' dhcp) (address 192.168.1.77) (netmask 255.255.255.0)
    (gateway (192.168.1.1)))))
 (ipv6 ())
 (ethernet
  ((method' auto) (interface wlp0s20f3) (address DC:FB:48:34:35:23)
   (mtu 1500)))
 (proxy
  (((host my-server) (port 1234)
    (credentials (((user my_user) (password <opaque>)))))))
 (nameservers ()))
```

The service when things go wrong:
```
((_proxy <opaque>) (_manager <opaque>)
 (id wifi_dcfb48343523_497474414d6f6b6b486f6c414d6f6b6b_managed_psk)
 (name IttAMokkHolAMokk) (type' Wifi) (state Configuration) (strength (90))
 (favorite true) (autoconnect true) (ipv4 ()) (ipv6 ())
 (ethernet
  ((method' auto) (interface wlp0s20f3) (address DC:FB:48:34:35:23)
   (mtu 1500)))
 (proxy ()) (nameservers ()))
```

The configuration is missing and is in the `Configuration` state whereas it should be `Online`.

Something this made me thing of:
When using networkmanager, I sometimes notice the connection dropping and reconnecting after a couple of seconds. I don't notice this when using connman, but it could be because connman connects much faster. I wonder if something is wrong with my network adapter.

guyonvarch:
I can’t think of a simple solution to this problem. More complex solutions involve storing the proxy outside of connman, or waiting for some time to let some time to the network to reconnect.

Suggesting to move forward with the failure solution for the moment.

controller/server/gui.ml
```diff
@@ -257,29 +257,28 @@ module NetworkGui = struct
...
-          | _ ->
-              (* No existing proxy credentials, use empty password *)
-              return (Some (Service.Proxy.make ~user:user ~password:"" host port))
-        )
+      let keep_password = 
+        form_data |> List.assoc_opt "keep_password" |> Option.is_some
+      in
+      let password = 
+        match (keep_password, current_proxy_opt) with
+        | (true, Some ({ credentials = Some { password } })) -> Ok (Some password)
```
knuton, line 265: This seems to introduce a vulnerability. It allows to change server while keeping the password unchanged, which allows revealing the password to a server an attacker controls.

guyonvarch: Good point.

We could check in the backend if the host (and port ?) didn’t change. 

Ideally, this should be reflected in the UI as well, but we may need a web component to generate the complete section, because there are interactions between elements: when changing the host, remove “keep password” and show a blank password input.

Maybe we should just remove this logic and let the user type the password if a modification is made ?

guyonvarch: > Maybe we should just remove this logic and let the user type the password if a modification is made ?

I forgot a bit how the previous mechanism worked, and may have misunderstood it a bit.

Password was kept only if proxy configuration was the same. This is useful in case someone wants to modify only the static IP configuration. Typing the password in the proxy form is not appropriate when someone wants to modify only the static IP configuration.

So, I think the previous solution was quite good, and only missed removing the password: add a “remove password” button to take care of this ?

PR #94, opened 2022-04-20
Improve font size coherence

guyonvarch:
This particularly improves the readability of proxy forms. Also remove some identifiers and wrap inside a label instead.

**Tip:** hide whitespaces to simplify the diff.

## Before

image

## Checklist

-   ~~[ ] Changelog updated~~
-   ~~[ ] Code documented~~


krksgbr (approved):
It looks good!

PR #95, opened 2022-04-21
Improve layout of the Controller

guyonvarch:
## Demo

https://user-images.githubusercontent.com/6768842/164464669-71d3f97a-bdbb-40f7-b438-3954dea10c04.mp4

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   ~~[ ] User manual updated~~ (if included in the release, will be done before)


controller/server/view/common/page.ml
```diff
@@ -55,28 +64,43 @@ let html ?current_page content =
...
+      :: header 
+      @ [ main
+              ~a:[ a_class [ "d-Layout__Main" ] ]
+              [ content ]
       ; script ~a:[ a_src "/static/client.js" ] (txt "")
-      ]
+      ])
     )
+
+let header_title ?back_url ?icon ?right_action content =
```
krksgbr, line 85: It would make better sense for me if this was called simply `header` as the `title` seems to be a child element of this.

krksgbr (changes requested):
**Unused styles**

I have found a few css classes in `style.css` that are not in used anymore:
- `Network__InternetConnect`
- `Network__InternetNotConnected`
- `Network__Title`
- `Network__SignalStrength`


**Orphan BEM elements**
I noticed we have quite a few BEM elements (ie. `d-Block__Element`) without a parent block (`d-Block`). I think if we have BEM elements without a corresponding block, we lose some clarity that BEM offers. For example if I see `d-Menu__Item`, then I'd also expect to find  a `d-Menu` to which I could make adjustments in case I want to for example apply change the layout. 

A few of such classnames already existed before this PR. It seems we've started using the `Block` part of BEM as simply name-spacing things, whereas a block should identify a self-contained (and potentially) re-usable entity. I think it would be useful to see if maybe some of these elements should be promoted from elements to blocks (ie. `MenuItem` instead of `Menu__Item`) - if they are self-contained, or to rename an ancestor of such an element to be a block - for example `d-Layout__Aside` seems a good candidate to be called `d-Menu`.

PR #96, opened 2022-04-28
Add check to avoid disclosing proxy passwords

knuton:
As PlayOS systems are often semi-public, secrets entrusted to the system should be protected from unwarranted access.

We should not allow users to keep the same proxy password when changing the rest of the proxy configuration, as this could be used to reveal the password to untrusted proxy servers on the same network or Internet.

This adds a check to fail update if keeping password was requested but other proxy fields were changed.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -262,7 +262,14 @@ module NetworkGui = struct
       in
       let password = 
         match (keep_password, current_proxy_opt) with
-        | (true, Some ({ credentials = Some { password } })) -> Ok (Some password)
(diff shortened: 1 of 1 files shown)


knuton:
@guyonvarch I opted to re-add the check if the keep-password checkbox is checked. This seems easier than reverting the commit that interacts with other UI changes, and the scenario of changing to an empty password continues to be supported.

guyonvarch (approved):
We could still improve the situation in the future if we want: when modifying host, port or user, if it’s different than before, uncheck preserve password box, disable it, and show the password field.

PR #104, opened 2022-10-27
Upgrade to nixpgs 22.05

guyonvarch:
## Testing

- [x] Standalone kiosk
- [x] Standalone controller
- [x] Run VM
- [x] Run live
- [x] Install on VirtualBox VM

## Checklist

-  [x] Changelog updated
-  ~~Code documented~~
-  ~~User manual updated~~


knuton:
Nice, looks good! I will test next week.

e8dfd95 seems to have a typo in the commit message?

knuton:
I don't get any sound from the kiosk (PC is connected to a TV via HDMI, including sound). Quick investigation: When logging in via remote maintenance mode I can make a horrible noise via `paplay < /dev/urandom`. When connecting to the kiosk with DevTools I can also see that the `AudioContext` is in `running` state in the browser. So the problem may be that the kiosk is not using the right sink or is otherwise silently failing to play audio.

---

Besides that I am seeing some errors that don't seem to cause an issue, but I'd like to check:

- When shutting down an installed system: `[FAILED] Failed unmounting /nix/store.`
- When booting an installed system: `systemd-gpt-auto-generator failed with exit status 1`

The unmounting failure should be benign, but it looks a bit unsettling and it would be good to suppress it if nothing else needs to be done.

When building I am also seeing these warnings that I don't think were there before:

```
trace: warning: system.stateVersion is not set, defaulting to 22.05. Read why this matters on https://nixos.org/manual/nixos/stable/options.html#opt-system.stateVersion.
```

This one should be simple to address.

knuton:
Skipping ahead to current Nixpkgs in #109.

PR #109, opened 2023-01-12
Upgrade to nixpkgs 22.11

knuton:
Upgrades nixpkgs to 22.11 channel. This builds on the work done by @guyonvarch to upgrade to nixpkgs 22.05 in #104.

Compared with [my comment](https://github.com/dividat/playos/pull/104#issuecomment-1308376792) on #104:

- Audio has been restored in cf34e0688201389a119f8c836a634ca87f993615
- The error `systemd-gpt-auto-generator failed with exit status 1` during boot has disappeared.

## Still to be resolved:

- <del>[ ] Error message when shutting down an installed system: `[FAILED] Failed unmounting /nix/store.`</del> Non-blocking, separate issue filed.
- [x] Harmless but distracting warning when building: `trace: warning: system.stateVersion is not set, defaulting to 22.05. Read why this matters on https://nixos.org/manual/nixos/stable/options.html#opt-system.stateVersion.`

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


knuton:
> I think we could use the same version as the NixOS version, as we test on a running system that after an upgrade, everything’s working as expected. If we would have a state that is incompatible with this new version, this would cause a problem.

I agree, that seems to be the best solution. Another solution might have been to keep this constant from the very first PlayOS release, but we didn't. Committed as 8f3f0f4dca110fc2cb89d3cf7bbf5e20ef0f7ce7.

I identified the sites triggering the warnings with `NIX_ABORT_ON_WARN=1 nix-build --show-trace` (need to remove the warning about the dummy cert in top level `default.nix`).

knuton:
- [x] Live system
  - [x] Wifi
  - [x] Basic system functionality (Play diagnostics) 
- [x] Installed system
  - [x] Wifi
  - [x] Localization
  - [x] RFID
  - [x] Senso
  - [x] Remote maintenance
  - [ ] Self update

PR #110, opened 2023-01-30
Include basic tools for network troubleshooting

knuton:
This makes `ethtool`, `dig`, `nslookup` and `traceroute` available system-wide. Even more basic tools such as `ping`, `netstat`/`ss` are already included.

We initially discussed this as "tiny binaries to include opportunistically". After considering various tools that _might_ be useful I concluded that opportunistic inclusion is not called for: Some tools that may be of great utility in some scenarios also allow for abuse to reconaissance or attack from inside a network.

The added tools are relatively limited in ability and their inclusion is motivated by common scenarios (bad Ethernet outlet or cable; DNS misconfiguration; problematic routing).

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


knuton:
> For `bind`, it seems it brings `dnssec-something`, `bind` being a command already available.

Thanks, I selected a more specific output for dnsutils instead.

> Would we need those tools as well in the rescue system ?

Right now I am thinking that we should remove the shell access in the rescue system entirely. Another possibility would be to bring it back in a more restricted form and we could consider including some analysis tools. However, this system would be a different system from the actual system normally booted, and it would not share its network configuration. So it's not entirely clear how useful network analysis tools would really be. My impulse right now is to just drop it.

PR #111, opened 2023-02-01
Recovery mechanism for status.ini

knuton:
This adds a mechanism to recover from emptied RAUC `status.ini` files if there is a leftover temp copy.

It also adds simple integration tests and makes them run on GitHub Actions.

A later effort might revise the partitioning in the installation script to avoid the problem entirely in new machines.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


krksgbr (approved):
All seems good to me!
I ran the test as well as tested that the recovery mechanism works in virtualbox.

PR #114, opened 2023-02-14
Enable quiet boot in application level config

knuton:
NixOS uses a very high loglevel default of 7, causing a very chatty startup process. We can reduce this by picking a more end-user friendly (calmer) level.

If more detail is needed in some cases, the system may be booted in rescue mode with fully chatty startup.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/bootloader/grub.cfg
+++ b/bootloader/grub.cfg
@@ -64,7 +64,7 @@ save_env a_TRY b_TRY
 search --label system.a --set a_DISK
 search --label system.b --set b_DISK 
 
-set common_kernel_args="$machine_id_arg quiet panic=3 boot.panic_on_fail"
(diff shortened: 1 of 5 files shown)


knuton:
I realized that this isn't covering all cases yet, need to revise it.

knuton:
With 96d436c this is covering all relevant cases.

However, I am not currently sure this is a good idea. As the loglevel is configured via the bootloader and boot partition, this configuration is not updated with the system partitions. So we can not easily go from one level to the other and different generation systems might then show different outputs.

There is no particular issue with the chatty output, so prefer not to go ahead with this for now.

PR #116, opened 2023-03-08
Add content-type to controller responses

knuton:
Include the content-type header to let sensitive clients (most TUI browsers) know what to do with the response.

I noticed this was missing while checking whether w3m/lynx etc. could be a lightweight option for using the controller in text-only systems.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


krksgbr (approved):
👍

PR #120, opened 2023-04-13
Replace DH110 setup screenshots with HDMI capture card shots

knuton:
Same content, slightly nicer look.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


knuton:
A cheap little USB dongle that takes HDMI as input and then acts as a web cam towards the PC it is plugged into.

PR #125, opened 2023-08-24
Add persistent storage stats in status console

knuton:
This allows to inspect the usage of the persistent partition, which is otherwise not inspectable without remote access. Other partitions are not included on the assumption that a reboot would fix everything due to their volatile state of being.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


krksgbr (approved):
👍

PR #129, opened 2023-09-05
Exclude installer medium as target

knuton:
Adds automatic exclusion of the installer medium from installation targets.

This helps prevent mysterious issues in cases such as

- atypically large USB drive as installer medium,
- missing or undetected HDD due to misconfiguration,

in which case the installer previously could end up trying to install to the USB stick it was booted from. This is bad in two ways, 1) it messes up the installer medium which needs to be re-flashed, 2) it is confusing and hides the issue as installation seems to work initially, but no system is installed upon reboot.

This also changes the script to list all devices it finds, as well as the devices it considers for installation. This should help understand issues with disks more quickly.

## Examples

Good disk found:

image

No good disk found (misconfigured BIOS):

image

In the latter case the installer would have previously installed to the USB drive.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -5,6 +5,7 @@
 - os: Widen rules for captive portal detection
 - os: Split system definition into base and application layers
 - os: Make build outputs and displayed system name configurable through application layer
+- os: Change installer script to exclude installer medium from installation targets
 - os: Include live system ISO in deployed outputs
 
 # [2023.2.0] - 2023-03-06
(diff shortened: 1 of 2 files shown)


knuton:
A limitation documented as a side-note in https://github.com/dividat/playos/pull/137#pullrequestreview-1620497744 is that when the installer medium is a CD (on VirtualBox), the current mechanism does not detect and exclude it:

image

PR #130, opened 2023-09-05
Make Full HD the default mode for PlayOS

knuton:
In the context of Play installations, PlayOS is almost always configured to run in "Full HD" mode, _especially_ on UHD screens.

We therefore change the default to enter Full HD mode. If a preference is set, it will continue to be used as before.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/application.nix
+++ b/application.nix
@@ -70,11 +70,18 @@ rec {
                 setxkbmap $(cat /var/lib/gui-localization/keymap) || true
               fi
 
-              # force resolution
-              scaling_pref=/var/lib/gui-localization/screen-scaling
-              if [ -f "$scaling_pref" ] && [ $(cat "$scaling_pref") = "full-hd" ]; then
-                 xrandr --size 1920x1080
-              fi
+              # Set preferred screen resolution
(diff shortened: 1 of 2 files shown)


stoeffel:
Change seems reasonable, but I don't know how to test this.

krksgbr:
> Change seems reasonable, but I don't know how to test this.

Indeed I don't think we offer enough guidance in regards to testing such changes in the README. I'll add some recommendations to it then maybe we can have a look together if they make sense. 

In this particular case I started by copying the relevant parts of the changed script into a file and tested it in isolation to make sure the logic is robust. Performing an additional integration test manually is also a good idea. This PR is sort of a special case that is best tested on a physical machine (since it's changing display settings), but most often we can test more simply in a VM or even just running the controller on our own machine.

krksgbr (approved):
All good! I tested the script in isolation and found it to be robust under exceptional circumstances such as invalid or missing settings.
Also tested on an actual PlayOS machine and it worked as expected.

PR #136, opened 2023-09-08
Improve guidance on development/testing

krksgbr:
Some improvements to documentation, prompted by @stoeffel's comment in https://github.com/dividat/playos/pull/130 as well as coming back to PlayOS after a long break and needing to remember how to do certain things.

## Checklist

-   ~[ ] Changelog updated~
-   ~[ ] Code documented~
-   ~[ ] User manual updated~

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -43,11 +43,29 @@ For example: `nix build --arg buildInstaller false --arg buildBundle false` will
 
 A virtual machine (with test instrumentation) can be started without any of the above builds.
 
-## Testing
+## Components
+
+### Controller
+
+The [controller](controller/) service orchestrates the system's self-update and acts as a configuration interface for the options exposed to the user. It may be run directly on a Linux host for development purposes.
+
+### Kiosk
+
+The [kiosk](kiosk/) browser is used in the default configuration of PlayOS to run any web application in a full screen kiosk. It can be run directly on most hosts for development purposes.
+
+
+## System Testing
+
+To test integrated portions of the PlayOS system, there are several options/levels available:
+
+- Running a single system partition with QEMU (fast, partial, isolated)
+- Running a full system inside a virtual machine such as VirtualBox (slow, simulated full, isolated)
+- Running a single system partition from a USB stick on a physical machine (medium fast, partial, isolated)
+- Running a full system on a physical machine (slow, full, isolated)
 
 ### QEMU VM
 
-Most changes to system configuration and/or the contoller can be tested in a virtual machine.
+Most changes to system configuration and/or the controller can be tested in a virtual machine.
 To create and run a VM, run:
 
 ```bash
@@ -62,38 +80,31 @@ See the output of `./result/bin/run-in-vm --help` for more information.
 
 #### Guest networking
 
-The default user-mode network stack is used to create a virtual Ethernet connection with bridged Internet access for the guest. If you find that the guest has a dysfunctional Internet connection, check your host's firewall settings. If using ConnMan, restart ConnMan service and try again.
-
+The default user-mode network stack is used to create a virtual Ethernet connection with bridged Internet access for the guest. If you find that the guest has a dysfunctional Internet connection, check your host's firewall settings. If using ConnMan on the host, restart the ConnMan service and try again.
 
 ### VirtualBox VM
 
-PlayOS can also be tested on a VM like VirtualBox, which can simulate a system more fully, including the installer, Grub and A/B partitions.
-Guidance for setting this up can be found [here](./docs/arch/Readme.org#installation-on-virtualbox).
+PlayOS can also be tested on a VM like VirtualBox, which can simulate a system more fully, including the installer, Grub and A/B partitions. Guidance for setting this up can be found [here](./docs/arch/Readme.org#installation-on-virtualbox).
 
 ### Testing on PlayOS hardware
 
-Changes such as NixOS upgrades, or to anything else that directly interacts with system hardware may necessitate testing on an actual machine.
-This can be done via booting from a live system or performing a complete installation.
+Changes such as NixOS upgrades, or to anything else that directly interacts with system hardware may necessitate testing on physical hardware. This can be done by booting from a live system or performing a complete installation.
 
-To build a live system:
+To build only a live system:
 
 ```bash
 nix-build --arg buildInstaller false --arg buildBundle false --arg buildDisk false
 ```
 
-To build an installer:
+To build only an installer:
 
 ```bash
 nix-build --arg buildLive false --arg buildBundle false --arg buildDisk false
 ```
 
-Flash the iso in `./result/` to a USB stick and boot or install PlayOS.
-Building a complete system like this is a lengthy process, so it is a good idea to test at the component level (controller, kiosk) first, where possible. 
+Flash the ISO in `./result/` to a USB stick and boot or install PlayOS.
 
-### Controller testing
-
-When making changes to the controller, it's possible to run and test it directly on your machine, without a VM or a live system. See the Controller's [Readme](controller/Readme.md) for details.
-
+Building a complete system takes time, so it is a good idea to test at the component or QEMU VM level first, where possible.
 
 ### Automated Testing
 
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -43,24 +43,44 @@ For example: `nix build --arg buildInstaller false --arg buildBundle false` will
...
+See the output of `./result/bin/run-in-vm --help` for more information.
 
 #### Guest networking
 
 The default user-mode network stack is used to create a virtual Ethernet connection with bridged Internet access for the guest. If you find that the guest has a dysfunctional Internet connection, check your host's firewall settings. If using ConnMan, restart ConnMan service and try again.
 
-## Testing
+### Testing on PlayOS hardware
+
+Changes such as NixOS upgrades, or to anything else that directly interacts with system hardware may necessitate testing on an actual machine. To do so, build a live system:
```
knuton, line 69: Two comments:

- There is a third possibility that sits between testing on QEMU and a physical machine: A [VM like VirtualBox](https://github.com/dividat/playos/tree/develop/docs/arch#installation-on-virtualbox), which can simulate a system more fully, including the installer, Grub and A/B partitions.
- The live system is very useful, but sometimes it's necessary to install a system.

Readme.md
```diff
@@ -43,24 +43,44 @@ For example: `nix build --arg buildInstaller false --arg buildBundle false` will
...
+
+```bash
+nix-build --arg buildInstaller false --arg buildBundle false --arg buildDisk false
+```
+
+Flash the iso in `./result/` to a USB stick and boot from this stick on the PlayOS computer.
+Unfortunately testing this way is currently a lengthy process and we would like to setup a workflow with a tighter feedback loop in the future.
```
knuton, line 76: I would leave out the promise of a better world from the docs. ;-)

We could say that building a complete system like this is a lengthy process, so it is a good idea to test at the component level (controller, kiosk) first, where possible. Totally for mentioning that this loop is arduous, would just track the intention to provide an alternative outside of the docs.

build
```diff
@@ -25,12 +25,20 @@ if [ "$TARGET" == "vm" ]; then
...
+echo -e "
+Run ./result/bin/run-in-vm to start a VM.
+
+QEMU Shortcuts:
+- Grab/release cursor: ctrl-alt-g
+- Quit: ctrl-alt-q
+
+VM shortcuts:
+Switch to QEMU monitor (ctrl-alt-2), activate shortcut via 'sendkey <some-shortcut>', then switch back to QEMU display (ctrl-alt-1).
+- TTY: ctrl-alt-f1
```
knuton, line 37: VT1 is normally disabled, though I was thinking about a debug mode (`playos.debug.enable = true;`) that enables virtual console 1, sets a known password for the root user, maybe adds tmux to system packages; it would have to refuse to compile under certain conditions, so it could never be left on by accident.

build
```diff
@@ -25,12 +25,20 @@ if [ "$TARGET" == "vm" ]; then
...
+- Status console: ctrl-alt-f8
+- Graphical UI: ctrl-alt-f7
+- Toggle controller: ctrl-shift-f12
```
knuton, line 40: These are not so much VM shortcuts, maybe move to a separate section?

The manual also explains these, should we link it more prominently?

controller/Readme.md
```diff
@@ -14,9 +14,99 @@ PlayOS controller is an OCaml application that manages various system tasks for
 - `server/`: main application code
 - `bin/`: binaries to start a dev server
 
+## Prerequisites
+Since PlayOS uses `Connman` for managing networks, it must be configured on your system.
+Replace your NetworkManager config with the following (remember to fill in `<your-host-name>`):
```
knuton, line 19: I have a much shorter config on my own machine (I use connman regularly/permanently), was this one necessary for you to use in full?

Maybe also good to note that this assumes NixOS, while it should be possible to build PlayOS on another Linux distro using just nix.

knuton (commented):
Thanks, good to write more of this down!

I do have some nitpicks. :nerd_face:

Readme.md
krksgbr: [Replace promise of a better future with suggestion](https://github.com/dividat/playos/pull/136/commits/48362bee24feb77381a431776c325ccd37c0d15c)

Readme.md
krksgbr: Mentioned these points in:

[Amend readme with additional details](https://github.com/dividat/playos/pull/136/commits/03340e1a0b37b7435071057191b02f382feebee4)

It isn't clearly explained when to use a live system vs a complete installation, because it's not really clear to me 😅. Not sure how important it is to go into such details. I would think the nature of a particular thing being tested would draw us to one of the options.

build
krksgbr: [Remove TTY shortcut hint as it is normally disabled](https://github.com/dividat/playos/pull/136/commits/3b812b49b8cb286c386ac699f9848418a33c50b6)

build
krksgbr: > These are not so much VM shortcuts, maybe move to a separate section?

I'm not sure I addressed this point in the way you meant, but I renamed the section:
[Rename shortcut section in VM hints](https://github.com/dividat/playos/pull/136/commits/1cb9932824d08f93be7073e69aa084709d192cd9)

Added a link to the manual:
[Link user manual in main readme](https://github.com/dividat/playos/pull/136/commits/df7f02286d7a82f574ec5034655be07fcade3ac8)

controller/Readme.md
krksgbr: I don't remember the details of why my config is so long, but it very well might contain unnecessary things. Could you share yours?

[Add note about assuming NixOS to Connman setup guidance](https://github.com/dividat/playos/pull/136/commits/5fd37cbfa209f3621d54096e7760af69b45adcad)

knuton: I think this should be enough (it is, I can use the controller like this):

```
  networking = {
    hostName = "foo";
    wireless.enable = true;
    networkmanager.enable = false;
  };
  
  services.connman = {
    enable = true;
    networkInterfaceBlacklist = [ "vboxnet" "zt" ];
  };
```

Readme.md
knuton: > Not sure how important it is to go into such details. I would think the nature of a particular thing being tested would draw us to one of the options.

I agree, I wanted to point these out because the added instructions initially were very specifically saying to use a live system, which could be misguiding at times.

> It isn't clearly explained when to use a live system vs a complete installation, because it's not really clear to me

As you say, an exhaustive characterization might be hard, but there are some obvious cases, based around the fact that the live system has no A/B partitions and entirely ephemeral storage. So:

- Testing anything to do with the installer requires a full system
- Testing anything to do with A/B systems requires a full system
- Testing anything that spans past one boot (e.g. "Full HD" settings) requires a full system

build
knuton: Toggling the controller works just fine for me with the regular key combination, also in QEMU. Does it not for you?

I was thinking to just move the sendkey explanation to the block above and let the PlayOS shortcut explanations stand on their own.

knuton:
@krksgbr Would you mind if I take you additions and revise the document on the basis of it?

I would push some changes and you can tell me whether the new explanations and structure match your intentions.

krksgbr:
> @krksgbr Would you mind if I take your additions and revise the document on the basis of it?
> 
> I would push some changes to this PR and you can tell me whether the new explanations and structure match your intentions.

Sure, feel free to do that.

Readme.md
knuton: 3d2526d

controller/Readme.md
knuton: https://github.com/dividat/playos/pull/136/commits/202d5eda1f640bc3697aaf6f218866338be714cf

build
knuton: https://github.com/dividat/playos/pull/136/commits/ac243d271b7636891b4d54bddc98a35ff422d33e

knuton (approved):
@krksgbr I left my edits, good for me now.

Your turn :)

krksgbr:
Thanks! Good for me as well.

PR #138, opened 2023-09-09
Pin only major version of Python to use binary cache

knuton:
Not all Python versions' packages are cached by Hydra, so using packages specifically from `python39` often means laboriously rebuilding packages with native extensions.

We can specify only that we want Python 3, the specific version this resolves to is going to be fixed as long as we are on the same nixpkgs pin.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -95,7 +95,7 @@ with pkgs; stdenv.mkDerivation {
 
   buildInputs = [
     rauc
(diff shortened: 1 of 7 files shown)


krksgbr (approved):
👌

PR #139, opened 2023-09-09
Fix fontconfig for weasyprint

knuton:
Override the weasyprint wrapper args to set our custom fontconfig again, adjusting for https://github.com/NixOS/nixpkgs/pull/207348.

## Checklist

-   [x] Changelog updated (fixes an unreleased issue with manual PDFs)
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -9,11 +9,16 @@ stdenv.mkDerivation {
 
   src = ./.;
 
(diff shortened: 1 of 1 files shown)


knuton:
@guyonvarch This one would be nice to land for the upcoming release, otherwise the manual PDF looks a bit sad. It only touches the docs, so the last pre-release functionality test remains valid.

PR #140, opened 2023-09-13
Fix controller dev-server

knuton:
python37Packages is no longer defined.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


knuton:
More seems to be wrong here, checking.

knuton:
@guyonvarch Thanks, this seems like a good suggestion.

Would you mind opening a PR with the change to watchexec? We can then close this one. I would go with the simple and slightly overeager invocation, the project compiles quickly enough. If it causes issues we can still use more specific rules.

PR #141, opened 2023-09-18
Fix controller dev-server

guyonvarch:
`python37Packages` is no longer defined in today's Nixpkgs, pywatchman does not work with Python 3.10[^1],

Use `watchexec`, that we already use in other places, which is simpler to use.

Additionally, we need to be more explicit about which inputs we need as packages in the shell.

[^1]: https://github.com/facebook/watchman/issues/970


knuton (approved):
Thanks

PR #142, opened 2023-11-07
Return no proxy instead of crashing if Connman is not installed

guyonvarch:
– With ConnMan installed, I should still be able to configure a manual proxy and use it,
– Without ConnMan on my system, the kiosk should start, and I should get no proxy.

See https://trello.com/c/tBj1L3t1/1594-playos-kiosk-return-none-proxy-if-there-is-no-connman-found-via-dbus

## Checklist

-   [ ] Changelog updated
-   [x] Code documented


knuton (approved):
:ok_hand: 

Verified that both branches are entered as expected and browser does not crash in either scenario.

PR #143, opened 2023-11-07
Add initial tests and typings for the Kiosk browser proxy

guyonvarch:
Noticing in https://github.com/dividat/playos/pull/142 that the code reading services was not so clear. This clarify this and also add unit tests and typing tests.


kiosk/kiosk_browser/proxy.py
```diff
@@ -1,4 +1,56 @@
-"""Monitor proxy changes and automatically apply changes in Qt application"""
+"""Monitor proxy changes and automatically apply changes in Qt application.
+
+A ConnMan service, given from Dbus, has the following type:
+
+type service = [ path, service_config ]
+
+type service_config = dict {
```
knuton, line 7: Would it make sense to actually use Python 3's typing features to write this as code instead of a comment?

kiosk/kiosk_browser/test_proxy.py
```diff
@@ -0,0 +1,21 @@
...
...efered', variant_level=1)}, signature=dbus.Signature('sv'), variant_level=1), dbus.String('Nameservers'): dbus.Array([], signature=dbus.Signature('s'), variant_level=1), dbus.String('Nameservers.Configuration'): dbus.Array([], signature=dbus.Signature('s'), variant_level=1), dbus.String('Timeservers'): dbus.Array([], signature=dbus.Signature('s'), variant_level=1), dbus.String('Timeservers.Configuration'): dbus.Array([], signature=dbus.Signature('s'), variant_level=1), dbus.String('Domains'): dbus.Array([], signature=dbus.Signature('s'), variant_level=1), dbus.String('Domains.Configuration'): dbus.Array([], signature=dbus.Signature('s'), variant_level=1), dbus.String('Proxy'): dbus.Dictionary({}, signature=dbus.Signature('sv'), variant_level=1), dbus.String('Proxy.Configuration'): dbus.Dictionary({dbus.String('Method'): dbus.String('direct', variant_level=1)}, signature=dbus.Signature('sv'), variant_level=1), dbus.String('mDNS'): dbus.Boolean(False, variant_level=1), dbus.String('mDNS.Configuration'): dbus.Boolean(False, variant_level=1), dbus.String('Provider'): dbus.Dictionary({}, signature=dbus.Signature('sv'), variant_level=1)}, signature=dbus.Signature('sv'))), signature=None)
```
knuton, line 6: OK to add them like this, but this is going to be hard to make sense of later.

As you just wrote it: Is there some sensible way to write the base boilerplate once and only override the parts of the service that matter for the test? A base struct and setting things on it? A function that takes parameters to then produce the required struct?

knuton (commented):
Looks good!

Some questions geared towards long term ease of maintenance.

stoeffel (approved):
Looks good to me!

PR #144, opened 2023-12-07
Use fake popups instead of system popups

guyonvarch:
Using popups, it was hard to test the kiosk without setting up PlayOS VM.

Aditionally:

- use only one webview,
- decorate as dialog when necessary,

There should be no visual changes.

See https://trello.com/c/SJNLN2OJ/1550-playos-dialog-less-settings-and-captive-portal

## Tests

- [x] Using local captive portal on VM, I can click on the message when settings are open.
- [x] With shed-key, use proxy of external server with basic auth.

```nix
networking.firewall.allowedTCPPorts = [ 1234 ];

services.tinyproxy = {
  enable = true;
  settings = {
    Port = 1234;
    Timeout = 600;
    Listen = "0.0.0.0";
    Allow = "12.13.87.my-ip";
    BasicAuth = "user password";
  };
};
```

## Checklist

-   ~~[ ] Changelog updated~~
-   [x] Code documented

### DIFF
--- a/kiosk/kiosk_browser/browser_widget.py
+++ b/kiosk/kiosk_browser/browser_widget.py
@@ -12,7 +12,7 @@ reload_on_network_error_after = 5000 # ms
 Webview loading status
 """
 class Status(Enum):
-    INITIAL_LOADING = auto()
+    LOADING = auto()
     NETWORK_ERROR = auto()
     LOADED = auto()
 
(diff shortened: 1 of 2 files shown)


knuton:
> [ ] Think about scenario with captive portal + proxy.

We should expect no difference, as [the proxy is set application wide](https://github.com/dividat/playos/blob/4a1a5bf3978c543f89934b29953af01e56ee4286/kiosk/kiosk_browser/proxy.py#L106-L112).

PR #145, opened 2023-12-08
Switch to main and release branches

guyonvarch:
This permits to simplify branch management.


build
```diff
@@ -3,20 +3,14 @@ set -euo pipefail
 
 TARGET="${1:-default}"
 
+scripts/info-branch-commit
+export readonly TARGET=$(scripts/option-prompt "Select a target" vm develop validation master lab-key shed-key stuck default)
```
knuton, line 7: Should we rename `default` to `all`?

knuton, line 7: I am not necessarily against an interactive script, but I wouldn't want that to be the only option, because

- I want to be able to script things,
- I want to be able to run a combined command like `./build vm; result/bin/run-in-vm` from history without having to type extra stuff into the terminal.

It would be fine for `./build` to prompt a target if none is given, I think?

knuton (approved):
Good for me now :+1:

PR #148, opened 2023-12-19
Add l10n options for Czech and Polish

knuton:
This uses the Polish Programmers keyboard, as it seems to be the most commonly used keymap despite its name[^1][^2].

I could not find as clear of an indication for Czech keyboard use. It might be there is a trend towards a QWERTY layout[^3], but overall the QWERTZ layout still seems more officially established.

[^1]: https://pl.wikipedia.org/wiki/Klawiatura_programisty
[^2]: https://en.wikipedia.org/wiki/List_of_QWERTY_keyboard_language_variants#Polish
[^3]: https://en.wikipedia.org/wiki/QWERTZ#Czech_(QWERTZ)

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -1,5 +1,9 @@
 # [UNRELEASED]
 
+## Added
+
+- os: Added localization options for Polish and Czech
+
 # [2023.9.0] - 2023-09-12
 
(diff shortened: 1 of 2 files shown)


stoeffel (approved):
LGTM!

PR #149, opened 2024-02-28
Focus webview when page is loaded

guyonvarch:
Tested on the kiosk only, when starting and already logged in, before the page has no focus, and with this change, it has focus.


knuton:
> Tested on the kiosk only, when starting and already logged in, before the page has no focus, and with this change, it has focus.

Great! What about when opening and closing configuration modal?

knuton (approved):

PR #150, opened 2024-02-29
[2023.9.0] Focus webview when page is loaded

guyonvarch:
Same as https://github.com/dividat/playos/pull/149, but targetting `release/23.9.x`.


knuton (approved):
Great, it's the oneliner we needed! :fist_raised: 

Confirming both the fresh boot and back-from-config scenario to result in focused app web view.

PR #155, opened 2024-03-31
HDMI hotplugging

knuton:
WIP

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/application.nix
+++ b/application.nix
@@ -25,35 +25,6 @@ rec {
     ];
 
     module = { config, lib, pkgs, ... }:
-    let
-      selectDisplay = pkgs.writeShellScriptBin "select-display" ''
-        # Discover connected displays
-        scaling_pref=$(cat /var/lib/gui-localization/screen-scaling 2>/dev/null || echo "default")
-        connected_outputs=$(${pkgs.xorg.xrandr}/bin/xrandr | ${pkgs.gnugrep}/bin/grep ' connected' | ${pkgs.gawk}/bin/awk '{ print $1 }')
-        if [ -z "$connected_outputs" ]; then
-          echo "No connected outputs found. Trying to apply xrandr globally."
-          case "$scaling_pref" in
-            "default" | "full-hd")
-              ${pkgs.xorg.xrandr}/bin/xrandr --size 1920x1080;;
-            "native")
-              ${pkgs.xorg.xrandr}/bin/xrandr --auto;;
-            *)
-              echo "Unknown scaling preference '$scaling_pref'. Ignoring.";;
-          esac
-        else
-          for output in $connected_outputs; do
-            case "$scaling_pref" in
-              "default" | "full-hd")
-                ${pkgs.xorg.xrandr}/bin/xrandr --output $output --mode 1920x1080;;
-              "native")
-                ${pkgs.xorg.xrandr}/bin/xrandr --output $output --auto;;
-              *)
-                echo "Unknown scaling preference '$scaling_pref'. Ignoring.";;
-            esac
-          done
-        fi
-      '';
-    in
     {
 
       imports = [ ./application/playos-status.nix ];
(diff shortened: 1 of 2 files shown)


knuton:
The current setup seems to cause the boot to hang, possibly the udev rule triggers while X is not yet up and running.

Alternative idea: Run the script as a one-shot systemd service, depend on `multi-user.target` (or something that guarantees X server is up), and restart the service from the udev rule when the event triggers.

knuton:
Closing this for now. We realized we also need to deal with the hotplugging in the kiosk, which currently has no notion of the resolution changing and needing to resize.

yfyf:
Reopening this, because it is a good starting point, just need to address a few extra points.

application.nix
```diff
@@ -126,6 +123,28 @@ rec {
...
+      # Monitor hotplugging
+      services.udev.extraRules = ''
+        ACTION=="change", SUBSYSTEM=="drm", RUN+="${pkgs.systemd}/bin/systemctl restart select-display.service"
+      '';
+      systemd.services."select-display" = {
+        description = "Select best display to output to";
+        serviceConfig.Type = "oneshot";
+        serviceConfig.ExecStart = "${pkgs.bash}/bin/bash ${./application/select-display.sh}";
+        serviceConfig.User = "play";
+        serviceConfig.After = "multi-user.target";
```
yfyf, line 135: I would expect this to be linked (through some rule) to the `graphical.target` instead? Also, I don't think the `After=`  will actually wait for the `multi-user.target`, so the `udev` rule will launch this too  early?

application.nix
```diff
@@ -126,6 +123,28 @@ rec {
...
+        environment = {
+          XAUTHORITY = "${config.users.users.play.home}/.Xauthority";
+          DISPLAY = ":0";
+        };
+        path = with pkgs; [
+          gnugrep
+          gawk
+          xorg.xrandr
+        ];
+        wantedBy = [ "multi-user.target" ];
```
yfyf, line 145: Ditto, probably best to remove it as @guyonvarch did in 85f5efa.

yfyf:
> We realized we also need to deal with the hotplugging in the kiosk, which currently has no notion of the resolution changing and needing to resize.

My proposal would be _not_ to do anything in the kiosk and simply restart it (or the whole xsession) after the screen is re-plugged? This is a one-off event that does not have to be handled very gracefully, but it would save us a lot of debugging when Qt fails to rescale. The only issue if the user hotplugs a screen mid-setup or mid-game with unsaved input, but that is kinda... unlikely?

application.nix
knuton: Honestly I'd need a moment to refresh my memory on the targets. I dimly remember trying to rely on a target that I then realized would not be present on our minimal system. But I just checked for `graphical.target` inside a VM PlayOS and it's there, so plausibly yes.

Regarding the ordering conditions, I understand `After=` to actually wait:

> After= ensures [...] that the listed unit is fully started up before the configured unit is started.

https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Before=

knuton:
> My proposal would be _not_ to do anything in the kiosk and simply restart it (or the whole xsession) after the screen is re-plugged? This is a one-off event that does not have to be handled very gracefully, but it would save us a lot of debugging when Qt fails to rescale. The only issue is if the user hotplugs a screen mid-setup or mid-game with unsaved input, but that is kinda... unlikely?
> 
> Edit: alternatively, Qt provides the [availableGeometryChanged](https://doc.qt.io/qt-6/qscreen.html#availableGeometry-prop) signal, but if we want to respond to it _without_ restarting, then we have propagate this to all child widgets, which might be non-trivial / hard to test?

My proposal would be to try with exactly the kind of mechanism you cite, which may or may not work well considering we have no proper window manager running. When we understand if/why it's hard to do, we can consider restarting the kiosk instead.

Restarting the kiosk/session may be problematic, if there is a problem with a cable/adapter/plug that causes it to reconnect at intervals.

application.nix
yfyf: > Regarding the ordering conditions, I understand After= to actually wait <..>

Yes, but only if both are being started at the same time, i.e. it is a "soft" dependency, from the same man page:

> If unit foo.service contains the setting Before=bar.service and **both units are being started**, bar.service's start-up is delayed until foo.service has finished starting up. After= is the inverse of Before=,

knuton: I see how you read it, though I wouldn't have read it like that.

What if we use both `After=` and `Requires=`?

yfyf:
Moving further development to #199

PR #157, opened 2024-04-30
Update driver for app-triggered firmware update support

knuton:
Update the driver to a new version supporting app-triggered firmware updates for Sensos.

The firewall is configured to allow the driver to act as a TFTP client, and the driver is configured to only allow requests from the kiosk URL's origin.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/application.nix
+++ b/application.nix
@@ -118,6 +118,17 @@ rec {
         };
       };
 
+      # Firewall configuration
+      networking.firewall = {
+        enable = true;
+
+        # Allow use of TFTP client for Senso firmware update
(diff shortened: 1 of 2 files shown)


knuton:
Looking for open UDP ports, `sudo nmap -v -sU -Pn -T4 <local-ip>` does not show any open ports:

```
Host is up (0.087s latency).
All 1000 scanned ports on <local-ip> are in ignored states.
Not shown: 1000 open|filtered udp ports (no-response)
```

PR #161, opened 2024-05-18
Fix permissible origin param for driver

knuton:
`ExecStart` may not use shell syntax, we need to invoke a shell in order to process the URL.

This was done incorrectly when added in
https://github.com/dividat/playos/pull/157, there must have been some confusion in our testing?

With the code as in #157, the driver runs, but does not accept requests from the Play application.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/application.nix
+++ b/application.nix
@@ -137,7 +137,7 @@ rec {
         description = "Dividat Driver";
         # Run driver with permissible origin limited to the origin of the kiosk URL
         serviceConfig.ExecStart = ''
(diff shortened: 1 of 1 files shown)


knuton:
> Though, how can I be sure the environment variable is correctly set? Would it be possible to log the permissible origin when starting the driver?

When going to the Play settings in the VM, you should see that it can not connect to the driver on `main`, but can connect here.

This does not strictly speaking tell you that the parameter is correctly set, because if no param is given, the permissible origin default range is going to include what is set for the VM's kiosk URL. You could force the kiosk to open Play from another URL, and then the driver should be unreachable again. The driver also logs rejected requests.

I guess, yes, we could also log the permissible origin when the driver starts up.

PR #165, opened 2024-05-24
Fix focus after reload

guyonvarch:
Since the migration to Qt6, reloading the browser, either using CTRL+R, or CTRL+SHIFT+R, the focus is lost on the webview.

By clearing the focus, it’s properly set afterward, after a reload.

I also tried to play with focus policy, and setting focus after a bit with a timer, but without success.

## Tests

- [x] Local kiosk
- [x] QEMU Vm


knuton (approved):
Thanks! Confirmed it makes the difference in QEMU VM.

PR #159, opened 2024-05-02
Allow using the kiosk on platforms without DBus

knuton:
We previously added support for using the kiosk without connman installed, but it still crashed when started on systems without DBus (i.e. any non-Linux platform).

With this change we add a stub implementation of the core proxy abstraction, which is going to provide safe no-op responses on non-Linux platforms, allowing to use the kiosk, albeit without proxy support.

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

PR #167, opened 2024-06-10
Kiosk development environment for macOS

krksgbr:
Follow up on https://github.com/dividat/playos/pull/159 to make the kiosk run without DBus.

The pyqt webengine bindings in nixpkgs are currently broken on darwin, so I created a separate development shell for macOS which makes use of [virtualenv](https://virtualenv.pypa.io/en/latest/) and [poetry](https://python-poetry.org/) to manage dependencies.

I'm requesting two reviews to test on both nixOS and macOS.

To run the kiosk on macOS:
```
cd kiosk
nix-shell ./macos/shell.nix
bin/kiosk-browser http://localhost:8080/play.html http://localhost:3333
```

## Checklist

-   ~[ ] Changelog updated~
-   [X] Code documented


.github/workflows/test.yml
```diff
@@ -43,6 +43,21 @@ jobs:
...
+    # run on GitHub's Ubuntu runners, as they don't support GUI applications.
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+    - uses: cachix/install-nix-action@v18
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
```
knuton, line 57: Maybe this can still also use the magic nix cache? Avoids rerunning the tests when no inputs have changed.

kiosk/kiosk_browser/main_widget.py
```diff
@@ -13,8 +15,17 @@ class MainWidget(QtWidgets.QWidget):
...
         super(MainWidget, self).__init__()
 
+        if platform.system() in ['Darwin']:
+            from kiosk_browser.proxy import Proxy
+            import os
+            system = System(name = "PlayOS",
+                            version = os.getenv("PLAYOS_VERSION","1.0.0-dev"))
+        else:
+            from kiosk_browser.dbus_proxy import DBusProxy as Proxy
+            system = System()
```
knuton, line 25: What if we moved this branching to the `system` module/`System` class?

It could offer something like a `def infer() -> System` function.

knuton (changes requested):
Looks good :+1: 

One small suggestion after code review. I am still going to perform practical tests.

kiosk/kiosk_browser/main_widget.py
krksgbr: Something like this?

[Move platform logic to system module](https://github.com/dividat/playos/pull/167/commits/4213269a2466a873fab377093013d222df494e44)

.github/workflows/test.yml
krksgbr: [Use magic-nix-cache for platform-smoke-test](https://github.com/dividat/playos/pull/167/commits/296217b1dc0ae2be2fbe6719fd3225cef0bca208)

stoeffel:
Yay! Works on my machine!

stoeffel (approved):
Works on my machine!


I have no context on the code, but the changes look reasonable to me.

knuton:
I seem to have found a problem on Linux, which is due to the Qt6 migration, not this PR. I am going to open a PR against the release branch. Otherwise this also looks good to me.

Looks like a one line fix, which we can then apply here.

knuton:
Superseded by https://github.com/dividat/playos/pull/279

PR #169, opened 2024-06-28
Revert Qt 6 migration of kiosk browser

knuton:
The Qt 6 version of QtWebEngine brought too many sneaky breaking changes and is holding up this PlayOS release because of it.

Revert to unblock releasing the other features and component updates. Migration can be picked up without having to rush it.

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated


knuton:
Some links related to the profile issue:

- https://forum.qt.io/topic/131277/qtwebengine-cannot-save-cache (ephemeral profile)
- https://github.com/qutebrowser/qutebrowser/issues/7202 (backwards compat)

PR #170, opened 2024-07-24
Update build/quick start instructions

knuton:
The quick start instructions were actually more like "slow start" instructions, suggesting to do a full build. Even worse, `nix build` fails in modern versions of the Nix CLI, which expects to find a flake file with this invocation.

This suggests to use the commonly used `./build` script instead and updates it to provide a minimal bit of guidance when invoking it empty (instead of doing, again, the full build).

A full build can still be done by invoking `./build all`.

For @yfyf.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -21,18 +21,28 @@ This repository contains the application layer for running the Dividat Play web 
 
 See the [documentation](docs/arch) and [user manual](docs/user-manual) for more information.
 
-## Quick start
+## Building the system
 
-Running `nix build` will create following (in `result/`):
+### Quick start
+
+The quickest way to see the system in action is to [use the QEMU VM](#qemu-vm).
+
+You can perform all common build variants using the `build` script. Run `./build` without parameters for a list of build variants.
+
+### Full build
+
+Running `./build all` (equivalently `nix-build`) creates all of the following (in `result/`):
 
 - `bin/`: Tools
 - `playos-VERSION-UNSIGNED.raucb`: RAUC bundle that can be used to update systems. Note that it is signed with a dummy development key. Real deployments would resign the bundle with `rauc resign`.
 - `playos-installer-VERSION.iso`: Bootable ISO image that can install the system.
 - `disk.img`: Preinstalled disk with bootloader, system partitions A/B and data partitions for testing (but without test instrumentation).
 
+Note that a full build takes a long time.
+
 ### Choose what to build
 
-For quicker development cycles you may pass following arguments to the build:
+The following arguments may be used to control the build outputs:
 
 - `buildInstaller`: Should the installer ISO image be built.
 - `buildBundle`: Should the RAUC bundle be built.
@@ -40,19 +50,6 @@ For quicker development cycles you may pass following arguments to the build:
 - `buildLive`: Should the PlayOS live system image be built.
 
 For example: `nix build --arg buildInstaller false --arg buildBundle false` will only build the system toplevels and the preinstalled disk image.
-
-A virtual machine (with test instrumentation) can be started without any of the above builds.
-
-## Components
-
-### Controller
-
-The [controller](controller/) service orchestrates the system's self-update and acts as a configuration interface for the options exposed to the user. It may be run directly on a Linux host for development purposes.
-
-### Kiosk
-
-The [kiosk](kiosk/) browser is used in the default configuration of PlayOS to run any web application in a full screen kiosk. It can be run directly on most hosts for development purposes.
-
 
 ## System Testing
 
@@ -138,6 +135,16 @@ To release an update to the `develop` channel:
 ### Key switch
 
 When switching key pairs on a channel, the new certficiate must be built into the bundle, which must then be signed with the old key. For this purpose, the `--override-cert` option of the deploy script is needed to provide RAUC with a certificate matching the new key.
+
+## Components
+
+### Controller
+
+The [controller](controller/) service orchestrates the system's self-update and acts as a configuration interface for the options exposed to the user. It may be run directly on a Linux host for development purposes.
+
+### Kiosk
+
+The [kiosk](kiosk/) browser is used in the default configuration of PlayOS to run any web application in a full screen kiosk. It can be run directly on most hosts for development purposes.
 
 ## Change Log
 
--- a/build
+++ b/build
@@ -1,7 +1,15 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-TARGET="${1:-default}"
+usage() {
+  echo "Usage: $0 {vm|develop|validation|master|stuck|lab-key|shed-key|all}"
+  exit 1
+}
+
+if [ $# -eq 0 ]; then
+  usage
+fi
+TARGET="$1"
 
 # Allow unfree packages, required for zerotier using a BSL 1.1 licence
 # See https://nixos.wiki/wiki/FAQ/How_can_I_install_a_proprietary_or_unfree_package%3F
@@ -106,7 +114,7 @@ elif [ "$TARGET" == "shed-key" ]; then
     --arg buildBundle false \
     --arg buildDisk false)
 
-elif [ "$TARGET" == "default" ]; then
+elif [ "$TARGET" == "all" ]; then
 
   (set -x; nix-build)
 
@@ -115,6 +123,6 @@ else
   echo
   echo "I do not know how to build '$TARGET'."
   echo "Aborting."
-  exit 1
+  usage
 
 fi


yfyf:
This is great, though `nix-build` / `./build all` fails for me with:
```
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'playos-2024.7.0-VALIDATION'
         whose name attribute is located at /nix/store/27rg15w6bjjh2hkbdqajhvqd7b02v5cb-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'buildCommand' of derivation 'playos-2024.7.0-VALIDATION'
         at /home/yfyf/src/dividat/playos/default.nix:102:3:
          101|
          102|   buildCommand = ''
             |   ^
          103|     mkdir -p $out

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: The option `programs.activationScripts' does not exist. Definition values:
       - In `/nix/store/27rg15w6bjjh2hkbdqajhvqd7b02v5cb-source/nixos/modules/services/x11/desktop-managers/cde.nix':
           {
             _type = "if";
             condition = false;
             content = {
               setup-cde = ''
           ...
```

knuton:
Odd, I do not get this.

Maybe related to running the build on a non-NixOS system somehow?

I pushed a commit to let the CI runner do a `./build all`, and it has the same issue you report: https://github.com/dividat/playos/actions/runs/10075744434/job/27854644214?pr=170

knuton:
Maybe you could check which of these args may be true for the build to work on your system?

https://github.com/dividat/playos/blob/95cd09b1d2268c11cb3ef7d8e0611236e6caa670/default.nix#L20-L23

`nix-build --arg buildInstaller false --arg buildBundle false --arg buildLive false --arg buildDisk true` etc. (this particular combination segfaults on GitHub's Ubuntu runner which is ... interesting)

yfyf:
The culprit seems to be `buildDisk` - settings to true causes the build to fail, though it throws a different error than `nix-build`:
```
$ export NIXPKGS_ALLOW_UNFREE=1 && nix-build --arg buildInstaller false --arg buildBundle false --arg buildLive false --arg buildDisk true
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'playos-2024.7.0-VALIDATION'
         whose name attribute is located at /nix/store/27rg15w6bjjh2hkbdqajhvqd7b02v5cb-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'buildCommand' of derivation 'playos-2024.7.0-VALIDATION'
         at /home/yfyf/src/dividat/playos/default.nix:102:3:
          101|
          102|   buildCommand = ''
             |   ^
          103|     mkdir -p $out

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: attribute 'config' missing
       at /nix/store/27rg15w6bjjh2hkbdqajhvqd7b02v5cb-source/lib/modules.nix:571:23:
          570|               #       options, so maybe that's why this is here.
          571|               isAttrs c.config || throw ''
             |                       ^
          572|                 In module `${c.file}', you're trying to define a value of type `${builtins.typeOf c.config}'
```

`nix-build` without any flags did also segfault earlier on on my machine _once_.

knuton:
My guess would be that the culprit is [`runInLinuxVM`](https://github.com/dividat/playos/blob/df62070342540bc1fad54b157f63c81bafa2b857/lib/make-disk-image.nix#L21C8-L21C29), for which there are similar sounding past issues:

- https://github.com/NixOS/nixpkgs/issues/73470
- https://github.com/NixOS/nixpkgs/issues/124371

I removed the CI test commits from this PR; the issue is unrelated.

knuton:
I think we could even consider removing this particular option: https://github.com/dividat/playos/commit/d871b01e3eafc3d1f3ea3e8c640ea58d393ce7ab

I have not really used it much for testing, myself. I maybe wouldn't remove it just now, we can wait for a bit more insight from a non-NixOS dev environment.

yfyf:
> Can you do `./build vm` and then run the VM?

Yes, `./build vm` and the built VM work totally fine!

yfyf:
@knuton returning to the issue of the failing `buildDisk` case. I am trying to fix it, because I want to repurpose it as the first stage of system-level testing.

There is something weird going on here and seems to be dependent on the evaluation order. Maybe you can spot the issue here, because I spent way too long trying to debug it and failed to find the root cause. Here's what I have figured out so far:

> My guess would be that the culprit is [`runInLinuxVM`](https://github.com/dividat/playos/blob/df62070342540bc1fad54b157f63c81bafa2b857/lib/make-disk-image.nix#L21C8-L21C29)

Nope! This has nothing to do with `runInLinuxVM`, because nix fails before even starting any real "building".

In particular, the [error cited above](https://github.com/dividat/playos/pull/170#issuecomment-2247849662) that I get when running `./build` with *only* `buildDisk=true` is due to some nix module (`c`) not having a `config` attribute. `--show-trace` reveals that the evaluation path is roughly `run-in-vm` -> `disk` -> `install-playos` -> `rescueSystem` -> `config.system.build.rescueSystem` (see [trace.txt](https://github.com/user-attachments/files/16680042/trace.txt) for the full trace).

Now, the weird thing is that **if I force `disk` to be built before `run-in-vm`** (e.g. by adding `ln -s ${components.disk} $out/disk.img` as the second line of `buildCommand`), **everything works**. The question is... why?

My guess is that something "bad" happens due to `bootloader/rescue` being [bizarrely recursive](https://github.com/dividat/playos/blob/26dc0346d813f4f38cf847e9d0e5134955f8c990/bootloader/rescue/default.nix#L57-L58) in how it [defines its build output](https://github.com/dividat/playos/blob/26dc0346d813f4f38cf847e9d0e5134955f8c990/bootloader/rescue/default.nix#L151-L161), though I might be wrong.

Anything else I could try to debug this further? Any way I could force `nix` to print the name of the mystery module `c` that is missing the `config` attribute? I know the "nix3" `nix build` has `--debugger` flag for that, but that is not available for the `nix-build`.

yfyf:
> Any way I could force `nix` to print the name of the mystery module `c` that is missing the `config` attribute?

I did a naughty thing and patched the source file in `/nix/store` to trace it, but the magic `c` is just the root package of nixos (`/nix/store/27rg15w6bjjh2hkbdqajhvqd7b02v5cb-source/nixos/default.nix`), which does not reveal much.

knuton:
> My guess is that something "bad" happens due to `bootloader/rescue` being [bizarrely recursive](https://github.com/dividat/playos/blob/26dc0346d813f4f38cf847e9d0e5134955f8c990/bootloader/rescue/default.nix#L57-L58) in how it [defines its build output](https://github.com/dividat/playos/blob/26dc0346d813f4f38cf847e9d0e5134955f8c990/bootloader/rescue/default.nix#L151-L161), though I might be wrong.

I think this might be a red herring. The [bizarre recursion is just part of NixOS's module system](https://nixos.wiki/wiki/NixOS_modules#Under_the_hood), unless I am missing a particular aspect you are pointing to. The invocation is [evaluating the modules and returning an attribute set](https://github.com/NixOS/nixpkgs/blob/master/nixos/default.nix) with the resolved `config`, which is going to include the `system.build.rescueSystem` output. I am not saying it's not bizarre, but I think it's just the regular mechanism.

knuton:
I want to try to reproduce, you said you're running Debian?

knuton:
> Yes, Debian stable ("bookworm"), nix version 2.23 IIRC (I don't have access to the machine now, will let you know specifically later).

Thanks. Multi-user or single-user (daemon or no daemon)?

yfyf:
> Thanks. Multi-user or single-user (daemon or no daemon)?

There ya go, full info:
```
[~]λ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.0-23-amd64, Debian GNU/Linux, 12 (bookworm), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.23.3`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

yfyf:
@knuton so you were right, the `buildDisk` issue seems to be caused by some regression in `nix`. I can buildDisk with `2.8.0`, `2.18.5` (stable) and `2.24.5` (latest), but `2.23.3` fails. I have rolled back to stable nix, which fixes the issue.

I guess it's not worth digging deeper, let's merge this PR and forget about it :)

PR #174, opened 2024-08-29
Add spatial navigation to controller page

knuton:
Adds spatial navigation via the focus-shift library.

Currently it is not possible to tick radio buttons/checkboxes, because this would require pressing `Space`. We could use the same method as in the apps to perform a synthetic click when `Enter` is pressed:

```js
document.addEventListener('keydown', (event) => {
    if (event.key === 'Enter') {
      const activeElement = document.activeElement;
      if (activeElement instanceof HTMLInputElement &&
          (activeElement.type === 'checkbox' || activeElement.type === 'radio')) {
          activeElement.click();
          event.preventDefault();
      }
    }
});
```

It doesn't make a practical difference right now, because one cannot do anything with the checkbox-unfolded sections (network settings, requiring inputs).

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -1,3 +1,9 @@
+# [UNRELEASED]
+
+## Added
+
+- controller: Enable spatial navigation using the arrow keys
+
 # [2024.7.0] - 2024-08-02
 
 # [2024.7.0-VALIDATION] - 2024-07-02
--- a/controller/dune
+++ b/controller/dune
@@ -4,6 +4,7 @@
   (Changelog.html as Changelog.html)
   (gui/reset.css as static/reset.css)
   (gui/style.css as static/style.css)
+  (gui/vendor/focus-shift.js as static/vendor/focus-shift.js)
   (gui/client.js as static/client.js)))
 
 ; Disable missing-record-field-pattern warnings (partial matching),
--- a/controller/gui/vendor/focus-shift.js
+++ b/controller/gui/vendor/focus-shift.js
@@ -1,0 +1,962 @@
+/**
+ * @overview focus-shift, library for spatial navigation with arrow keys
+ *
+ * https://github.com/dividat/focus-shift
+ *
+ * @copyright Dividat AG, 2024
+ * @license MIT
+ */
+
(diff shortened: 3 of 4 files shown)


knuton:
> We don’t have a package manager to deal with the _focus-shift_ dependency, I think that’s OK. But I couldn’t find the version.

It's in the commit: https://github.com/dividat/playos/pull/174/commits/e2a2dbed6da1544cdd81dd980a4d1ea5aa604acb

knuton:
> I can’t update select inputs, I need to press _Space_:

That's interesting, it worked for me. Which browser is that with?

knuton:
> OK, though it would be more visible in the filename, or in the header of the file, more in the face.

Agree that in your face is good, but have an uneasy feeling with header (because it requires editing the file) and filename (SHA fragments are not self-documenting, and we don't currently have tagged versions of focus-shift -- maybe that's the problem). Wonder if we could use Nix to bring in the dependency instead.

Would leave as is for now. It's not ideal, but it's low overhead and seems manageable as long as there is one single JS dependency we control.

I can explore using Nix (or Bower? :smile:) if you think the commit message is too out of sight.

> With firefox. But testing using the kiosk, which is the target browser, it works as expected.

I checked, we have the same issue with Firefox in Play.

We can look into extending the glue to map Enter to Space a click there, and we can also use it here. I am still not convinced this glue should be part of the focus-shift library -- though at least in the way we use it, we would always want both the spatial navigation and the extended Enter key support.

> On the _Print label_ page, we can access any field, and also increase and decrease labels, but we sometimes have to find a path:

We can just drop that part of the controller, I think.

> Wondering if _Shutdown_ could be part of the list, so that going up from _Shutdown_ would go to Changelog.

Yes, it can, thanks! 5b5257a

knuton:
> > On the _Print label_ page, we can access any field, and also increase and decrease labels, but we sometimes have to find a path:
> 
> We can just drop that part of the controller, I think.

https://github.com/dividat/playos/pull/175

PR #176, opened 2024-09-19
Use tagged version of focus-shift

guyonvarch:
This indicates more in the face which version of focus-shift we are using.

This also upgrade `focus-shift` with the following change:

- Ignore non-rendered elements

We could think about using `nix` to pull the dependency, but as this may not be updated that much, and as we have only this dependency, keep it simple for now?


controller/dune
```diff
@@ -4,7 +4,7 @@
   (Changelog.html as Changelog.html)
   (gui/reset.css as static/reset.css)
   (gui/style.css as static/style.css)
-  (gui/vendor/focus-shift.js as static/vendor/focus-shift.js)
+  (gui/vendor/focus-shift-1-0-0.js as static/vendor/focus-shift.js)
```
knuton, line 7: Is there an issue with using the version number in the filename, `1.0.0`?

knuton: Yes, exactly. I think this is the conventional way, and therefore more easily recognized as a version number.

We use this same naming pattern for the ISOs here: https://github.com/dividat/playos/blob/5f6bfb694ff1e00ba824baac8f34314e79caef65/default.nix#L115-L120

PR #178, opened 2024-09-24
Change active VTE restriction implementation to a xmodmap script

yfyf:
⚠️🐃 yak-shaving alert! 🐃 ⚠️

It was very annoying to wait for xorg to be rebuilt just to make this small restriction, so I had to shave this yak. Perhaps not the most elegant approach, but it works and avoid needless re-compilation. Easier to maintain and more "pluggable" as well.

Manually tested via `run-in-vm` and QEMU's sendkey. Ctrl+Shift+F12 continues to work. As before, Ctrl+Alt+F8 + Ctrl-Alt+F{N} is a hack to "escape" the restriction.

This could actually be limited at the VTE/kernel level as well in a similar way using something like

        dumpkeys | grep .. Console_{N} .. | sed .. | loadkeys

but not sure we want that.

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated


knuton (approved):
Nice :ok_hand: 

Verified that VT1 can not be accessed directly, VT8 can, VT7 can. (Also, VT1 can from VT8, as before.)

PR #179, opened 2024-09-24
Open settings with a long press on the Menu key

guyonvarch:
## Demo

https://github.com/user-attachments/assets/6eb2eef9-0b69-418c-874c-9c05e875722a

Using https://github.com/dividat/diviapps/pull/2161 locally, pressing the Menu key with a short press first, and then pressing the Menu key with a long press.

## Limitations

**Fixed by** https://github.com/dividat/playos/pull/179/commits/891c1d5df5020dd9b4582947ac481ff90658c734.

Keeping the long press longer after the settings have been shown, it will trigger again after the delay the settings toggle. For some reason, we get a release signal, followed by a Press signal, after we toggle the settings.

https://github.com/user-attachments/assets/274b3c99-da37-473d-b315-892cc9fae4d5

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -2,6 +2,7 @@
 
 ## Added
 
+- kiosk: Open settings with a long press on the Menu key
 - controller: Enable spatial navigation using the arrow keys
 
 ## Removed
--- a/docs/user-manual/Readme.org
+++ b/docs/user-manual/Readme.org
@@ -33,7 +33,7 @@ When a remote control with a power button is connected, the system may be shut d
 
 * Administration
 
-<<administration>>A menu for system administration may be accessed with the key combination ~Ctrl-Shift-F12~.
+<<administration>>A menu for system administration may be accessed with the key combination ~Ctrl-Shift-F12~, or with a long press on the Menu key.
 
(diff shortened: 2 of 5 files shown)


knuton:
> Keeping the long press longer after the settings have been shown, it will trigger again after the delay the settings toggle. For some reason, we get a release signal, followed by a Press signal, after we toggle the settings.

Not sure I understand the problem yet, but maybe detect and ignore auto repeats? https://doc.qt.io/qt-5/qkeyevent.html#isAutoRepeat

knuton:
This change brings another side-effect: When pressing Ctrl-C in the terminal from which I launched the kiosk browser, I now get what looks like a crash.

```
^CTraceback (most recent call last):
  File "/home/emerij/Development/playos/kiosk/kiosk_browser/main_widget.py", line 85, in eventFilter
    def eventFilter(self, source, event):

KeyboardInterrupt
Aborted (core dumped)
```

Previously the Ctrl-C was ignored.

knuton:
I think we could also change it to exiting, but gracefully instead of crashing.

knuton (approved):
:zap:

PR #180, opened 2024-09-24
Move all controller deps to buildInputs, remove utop, nodejs

yfyf:
Having them in `propagatedBuildInputs` resulted in ocaml being included in the system image and bloated it by an extra 500 MB.

cc @knuton 

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated


knuton (approved):
:sparkles:

PR #182, opened 2024-09-24
Improve install device selection in installer

yfyf:
- Look for "/iso" mount not only in children of blockdevices, but top-level devices too (e.g. this happens when running qemu with `-cdrom playos-installer-<..>.iso`)
- Eliminate devices that are too small or read-only from potential installation targets.

cc @knuton 

## Checklist

-   [X] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -3,6 +3,7 @@
 ## Added
 
 - controller: Enable spatial navigation using the arrow keys
+- os: Improve installation device selection
 
 ## Removed
 
--- a/installer/install-playos/install-playos.py
+++ b/installer/install-playos/install-playos.py
@@ -14,6 +14,7 @@ from datetime import datetime
 
 PARTITION_SIZE_GB_SYSTEM = 10
 PARTITION_SIZE_GB_DATA = 5
+REQUIRED_INSTALL_DEVICE_SIZE_GB = PARTITION_SIZE_GB_DATA + 2*PARTITION_SIZE_GB_SYSTEM + 1
 
 GRUB_CFG = "@grubCfg@"
 GRUB_ENV = '/mnt/boot/grub/grubenv'
@@ -40,10 +41,20 @@ def find_device(device_path):
     boot_device = get_blockdevice("/iso")
     if boot_device is None:
         print("Could not identify installer medium. Considering all disks as installation targets.")
-        available_devices = all_devices
-    else:
-        available_devices = [device for device in all_devices if not device.path.startswith(boot_device)]
 
+    def device_filter(dev):
+        is_boot_device = (boot_device is not None) and dev.path.startswith(boot_device)
+        is_read_only = dev.readOnly
+        is_big_enough = gb_size(dev) > REQUIRED_INSTALL_DEVICE_SIZE_GB
+
+        return (not is_boot_device) and (not is_read_only) and is_big_enough
+
+    available_devices = [device for device in all_devices if device_filter(device)]
+
+    print("Minimum required size for installation target: {req_gb} GB".format(
+        req_gb=REQUIRED_INSTALL_DEVICE_SIZE_GB
+    ))
+    
     print(f"Found {len(available_devices)} possible installation targets:")
     for device in available_devices:
         print(f'\t{device_info(device)}')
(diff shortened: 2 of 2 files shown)


yfyf:
P.S. I was "inspired" to do this PR because I was running qemu with `-cdrom playos-installer-XX.iso` and it would fail to identify the mounted ISO as the boot medium (and hence would not filter it out from potential installation targets).

controller/Changelog.md
```diff
@@ -3,6 +3,7 @@
 ## Added
 
 - controller: Enable spatial navigation using the arrow keys
+- os: Improve install device selection
```
knuton, line 6:
```suggestion
- os: Improve installation device selection
```

installer/install-playos/install-playos.py
```diff
@@ -40,9 +41,15 @@ def find_device(device_path):
     boot_device = get_blockdevice("/iso")
     if boot_device is None:
         print("Could not identify installer medium. Considering all disks as installation targets.")
-        available_devices = all_devices
-    else:
-        available_devices = [device for device in all_devices if not device.path.startswith(boot_device)]
+
+    def device_filter(dev):
+        is_boot_device = (boot_device) and (dev.path.startswith(boot_device))
```
knuton, line 46: Maybe by being a bit more explicit it's clearer that all devices pass this if the boot device could not be identified:

```suggestion
        is_boot_device = (boot_device is not None) and (dev.path.startswith(boot_device))
```

installer/install-playos/install-playos.py
```diff
@@ -14,6 +14,7 @@
 
 PARTITION_SIZE_GB_SYSTEM = 10
 PARTITION_SIZE_GB_DATA = 5
+REQUIRED_INSTALL_DEVICE_SIZE_GB = PARTITION_SIZE_GB_DATA + PARTITION_SIZE_GB_SYSTEM + 1
```
knuton, line 17: We need room for two system partitions, right?

installer/install-playos/install-playos.py
```diff
@@ -40,9 +41,15 @@ def find_device(device_path):
...
+    def device_filter(dev):
+        is_boot_device = (boot_device) and (dev.path.startswith(boot_device))
+        is_read_only = dev.readOnly
+        is_big_enough = gb_size(dev) > REQUIRED_INSTALL_DEVICE_SIZE_GB
+
+        return (not is_boot_device) and (not is_read_only) and (is_big_enough)
+
+    available_devices = [device for device in all_devices if device_filter(device)]
 
     print(f"Found {len(available_devices)} possible installation targets:")
```
knuton, line 54: Maybe before this we could print the required disk space, so it's clear why certain devices listed above are not listed below.

knuton (changes requested):
Looks good! A few minor suggestions.

installer/install-playos/install-playos.py
```diff
@@ -40,9 +41,15 @@ def find_device(device_path):
...
-        available_devices = all_devices
-    else:
-        available_devices = [device for device in all_devices if not device.path.startswith(boot_device)]
+
+    def device_filter(dev):
+        is_boot_device = (boot_device) and (dev.path.startswith(boot_device))
+        is_read_only = dev.readOnly
+        is_big_enough = gb_size(dev) > REQUIRED_INSTALL_DEVICE_SIZE_GB
+
+        return (not is_boot_device) and (not is_read_only) and (is_big_enough)
```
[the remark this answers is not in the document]
yfyf: To be honest, I just did it without thinking, I guess to keep it symmetric, but it is slightly confusing, will remove.

yfyf:
Resolved all the comments, tested manually by adding additional disks and checking that it picks the right one based on size:

```
Found 2 disk devices:
        /dev/vdb (Virtio Block Device - 1 GB)
        /dev/vda (Virtio Block Device - 28 GB)
Could not identify installer medium. Considering all disks as installation targets.
Minimum required size for installation target: 26 GB
Found 1 possible installation targets:
        /dev/vda (Virtio Block Device - 28 GB)
```

(Installer medium is not identified because I am running via nix `runInLinuxVM`.

knuton:
> tested manually by adding additional disks and checking that it picks the right one based on size

Did the following additional sanity checks on physical machine:

- The 32 GB installation medium is ignored
- Additional 8 GB device is ignored
- 120 GB disk is selected

knuton (approved):
Thanks :ok_hand:

PR #186, opened 2024-10-14
Add switch, reboot and factory reset options / action calls to status page

yfyf:
This adds buttons / action calls to the controller's status page for:
- A button for switching to a newly installed system (if there is one) via a reboot
- A button for switching to a newer system when running from a (manually selected) older system
- A button for doing a factory reset by wiping persistent data (always displayed)
- A notice for doing a system reinstall when it appears to be borked

An integration test is added to check that the button actions perform the required actions.

An end-to-end test is added for the systemd wipe service. Could extend it to test all the scenarios, but it would be kinda messy / lengthy.

Several pre-requisite changes were also done, see commit-by-commit for details. 

Open questions:
- Should we also display "Switch to older system"? Might be useful to easily switch to older version after a broken update.
- How do you test Controller GUI besides manual testing?

TODOs:
- Changelog and user manual needs to be updated, but I am waiting on review on the actual final feature set before this. 

## Checklist

-   [X] Changelog updated
-   [X] Code documented
-   [X] User manual updated


testing/end-to-end/tests/base/factory-reset.nix
```diff
@@ -0,0 +1,43 @@
...
+
+with TestCase("Persistent data remains after reboot"):
+    playos.succeed("echo TEST_DATA > /mnt/data/persist-me")
+    playos.shutdown()
+    playos.start(allow_reboot=True)
+    playos.wait_for_unit('mnt-data.mount')
+    playos.succeed("grep TEST_DATA /mnt/data/persist-me")
+
+with TestCase("Persistent data is wiped if factory reset is triggered"):
+    playos.succeed("systemctl start playos-wipe-persistent-data.service")
```
knuton, line 36: I guess doing this via an actual HTTP request must fail somehow, or could we do this instead?

knuton (commented):
> An integration test is added to check that the button actions perform the required actions.
> An end-to-end test is added for the systemd wipe service. Could extend it to test all the scenarios, but it would be kinda messy / lengthy.

Can you elaborate on the messy/lengthy spectrum, more on the messy or lengthy side?

The advantage would be that we may not need to assume as much about the mechanism by which things happen.

> Should we also display "Switch to older system"? Might be useful to easily switch to older version after a broken update.

In this case I guess we could consider simply making this a "reboot into other system" button. Is there a reason _not_ to show the button, when both versions are the same? It could be a system toggle button, with maybe the label text changing according to the version gradient.

> How do you test Controller GUI besides manual testing?

We don't have a test suite for the controller GUI, it being/having been very minimal.

testing/end-to-end/tests/base/factory-reset.nix
yfyf: This absolutely can be done using an HTTP request, but I on purpose wanted to split the tests into two parts: testing the `playos-wipe-persistent-data.service` only (because it is a very tricky thing, took me ages to get it right!) and testing the REST endpoints of the controller GUI.

controller/gui/style.css
```diff
@@ -91,6 +91,12 @@ html {
     white-space: nowrap;
 }
 
+.d-Menu_Item--Shutdown {
+    cursor: pointer;
+    border: 0;
+    width: 100%;
+}
```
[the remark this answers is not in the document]
yfyf: You are probably right, I have no idea, I suck at CSS and was just winging this :grin:

yfyf:
> Can you elaborate on the messy/lengthy spectrum, more on the messy or lengthy side?

It will require browser automation using `pyppeteer`, mostly due to that. It is also probably not an optimal strategy to rely on e2e tests for every feature, because it will create a very "heavy" and difficult to maintain test suite. As we discussed earlier, it's a good idea to restrict e2e test usage to "core" functionality. 

I think a better plan would be, once #173 is merged, to re-use the mock RAUC and mock Update Client to create unit tests for the front-end, because then we can straightforwardly mock the backend for different scenarios? Alternatively, I could add add browser automation into the integration test, rather than the e2e test.

> The advantage would be that we may not need to assume as much about the mechanism by which things happen.

True, I guess it's mostly a long term test maintainability question. I am happy to implement the e2e tests for this feature set, if you think it is the best path forward.

> In this case I guess we could consider simply making this a "reboot into other system" button. Is there a reason _not_ to show the button, when both versions are the same? It could be a system toggle button, with maybe the label text changing according to the version gradient.

Agree, good idea, I will implement it as an (always-visible) toggle button, with different notice/label text based on the slot states.

testing/end-to-end/tests/base/factory-reset.nix
yfyf: Simplified in 5c8823880a2

controller/gui/style.css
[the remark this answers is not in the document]
yfyf: Simplified in 5c8823880a2

yfyf:
> Agree, good idea, I will implement it as an (always-visible) toggle button, with different notice/label text based on the slot states.

Ended up doing something slightly different in a7e6a5d. After considering more, I think an always-visible toggle button is not a good idea, because there are various cases where the user should not be permitted to switch (e.g. during or after update/installation, when RAUC produces an error, etc), hence an implementation that only offers the switch on specific Update states.

One more open question is whether switching to other slot should also be prohibited when non-booted slot status is marked as "bad"?

testing/integration/controller-system-buttons.nix
```diff
@@ -0,0 +1,131 @@
...
+
+    # Ensure persistent-data-wipe service is running
+    with subtest("Factory reset works"):
+        playos.succeed("""
+            systemctl show --no-pager \
+                --property ActiveState \
+                playos-wipe-persistent-data.service | grep 'ActiveState=inactive'
+        """)
+        playos.succeed("curl -X POST http://localhost:3333/system/factory-reset >&2")
+        playos.wait_for_console_text("systemd.*Starting playos-wipe-persistent-data.service.")
```
knuton, line 110: Maybe known, but this currently seems to fail, probably due to

`Error: # [   10.094713] playos-controller[726]: playos-controller: [ERROR] GUI Error: org.freedesktop.systemd1.NoSuchUnit: Unit boot.mount not found.`

knuton (commented):
Great, this seems simpler than the previous shutdown approach for wiping. :ok_hand:

testing/integration/factory-reset.nix
```diff
@@ -0,0 +1,93 @@
...
+            "/nix/store" = {
+              device = "nix-store";
+              fsType = "9p";
+              neededForBoot = true;
+              options = [
+                "trans=virtio" "version=9p2000.L"
+                "msize=${toString config.virtualisation.msize}"
+                "cache=loose"
+              ];
+            };
```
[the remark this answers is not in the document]
yfyf: Are you referring to the `fileSystems."/nix/store"` entry specifically? Then the answer is "no", because [qemu-vm.nix overrides all `fileSystems` with `virtualisation.fileSystems`](https://github.com/NixOS/nixpkgs/blob/96fd503b3a7856ed964e4c81e9ba89dbc81367ee/nixos/modules/virtualisation/qemu-vm.nix#L1163) unless it is unset (like here), but that implies you have to set up everything manually.

controller/server/view/status_page.ml
```diff
@@ -1,24 +1,131 @@
...
+            "Switch to other slot and reboot"
+      ]
+
+let factory_reset_call =
+    let confirm_msg = "This will wipe all configuration and user data. Are you sure?"
+    in
+      [ note "WARNING: Clears all user data and reboots the machine, resulting in a
+              fresh install state. Will require to manually reconfigure network,
+              localization and all other settings. Users will have to re-login
+              to Dividat Play."
```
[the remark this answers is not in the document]
yfyf: Rephrased in 02818b15ce5

testing/integration/controller-system-buttons.nix
yfyf: Fixed by c8082c988 and 71f2086de3

yfyf:
controller-system-buttons test was flaky because it was expecting messages from systemd-logind, but it was racing with controller on startup. 648328a should fix it, ran the test 10x without failures.

controller/server/view/status_page.ml
```diff
@@ -1,24 +1,131 @@
...
+        | (RebootRequired, _) ->
+            Some (Definition.description reboot_call)
+        | (OutOfDateVersionSelected, Status _) ->
+            Some (Definition.description (
+                switch_to_newer_system_call target_slot
+            ))
+        | (UpToDate _, Status _) ->
+            Some (Definition.description (
+                switch_to_older_system_call target_slot
+            ))
```
knuton, line 105: This is slightly misleading: `(UpToDate _, Status _)` matches a new installation, with both systems the same version.

In this case the message is a bit misleading:

image

Suppress the reboot option if the version numbers are identical?

knuton:
Nice, I think we're pretty much there!

If you do a rebase with `main` (combat flakiness) and address the one UI point, I'll go through some final manual scenarios.

controller/server/view/status_page.ml
yfyf: Good catch! Suppressed via 02bd61a

yfyf:
> If you do a rebase with `main` (combat flakiness) and address the one UI point, I'll go through some final manual scenarios.

Done!

knuton:
Manual scenarios:

- [x] Factory reset wipes stored config
- [ ] Switch to older system
- [ ] Switch to newer system
- [x] Reboot upon installation (`RebootRequired`)
- [ ] Other states
  - [x] `ErrorGettingVersionInfo`
  - [x] `Downloading`
  - [x] `Installing`
  - [ ] `UpToDate` with identical versions on system partitions

controller/server/view/status_page.ml
```diff
@@ -1,24 +1,131 @@
...
+let suggested_action_of_state (update:Update.state) (rauc:rauc_state) booted_slot =
+    let target_slot = other_slot booted_slot in
+    match (update, rauc) with
+        | (RebootRequired, _) ->
+            Some (Definition.description reboot_call)
+        | (OutOfDateVersionSelected, Status _) ->
+            Some (Definition.description (
+                switch_to_newer_system_call target_slot
+            ))
+        | (UpToDate {booted; inactive}, Status _) when booted != inactive ->
```
knuton, line 102:
```suggestion
        | (UpToDate {booted; inactive}, Status _) when booted <> inactive ->
```

Oops, missed this before, need to use [structural equality op](https://ocaml.org/manual/5.2/expr.html#ss%3Aexpr-operators).

knuton:
Small issue, but I think it's safe to update the manual now!

yfyf:
>  Small issue, but I think it's safe to update the manual now!

Done, updated the manual (7273d32) and changelog (06a588a).

Small nit-pick: was trying to exclude the level 3 header from ToC with `#+OPTIONS: toc:2`, but seems like `pandoc` does not support it?

knuton:
> Small nit-pick: was trying to exclude the level 3 header from ToC with `#+OPTIONS: toc:2`, but seems like `pandoc` does not support it?

Can not find clear answer, but possibly not. Just tell pandoc directly?

```diff
diff --git a/docs/default.nix b/docs/default.nix
index 0a3e389..2f128fb 100644
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -39,7 +39,7 @@ stdenv.mkDerivation {
     cd ../user-manual
     pandoc \
         --template ../templates/default.html \
-        --toc --number-sections \
+        --toc --number-sections --toc-depth=2 \
         -V version=${version} -M date=$DATE \
         -t html5 \
         --standalone --self-contained -o $out/user-manual.html \

```

knuton (approved):
Great!

Manual testing through a combination of real world scenarios and `nano /boot/status.ini; systemctl restart rauc.service; systemctl restart playos-controller.service` manipulations. Everything goes together well now.

PR #187, opened 2024-10-14
Ensure KVM is usable and add e2e tests in GH workflow 

yfyf:
This is a PoC that e2e tests can be made to run as part of GH workflow, but:
- It cannot be cached, because it will evict everything else from the GH action cache which [has a quota of 10GB](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).
- We cannot use a matrix workflow, because each runner will have to rebuild the disk, so tests will take forever. 
- It kinda pushes the GH runner storage to the limit.

@knuton thoughts?

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,3 +56,30 @@ jobs:
           system-features = nixos-test benchmark big-parallel kvm
     - uses: DeterminateSystems/magic-nix-cache-action@v8
     - run: NIXPKGS_ALLOW_UNFREE=1 ./build vm
+
+  e2e-tests:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Make more space available on the runner
+      run: |
+        sudo rm -rf /usr/share/dotnet \
+                    /usr/local/lib/android \
+                    /opt/ghc \
+                    /opt/hostedtoolcache/CodeQL
+
+    - name: Ensure KVM is usable by nix-build
+      run: sudo chmod a+rwx /dev/kvm
+
+    - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


knuton:
Thanks!

> * It cannot be cached, because it will evict everything else from the GH action cache which [has a quota of 10GB](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).

Looks like we can not change this right now, because there is no "cash for cache" scheme. It may be in the works: https://github.com/orgs/community/discussions/42506#discussioncomment-10388640

An alternative might be Cachix or some other external artifact cache. Of course we might likewise rip through the available storage quickly.

> * We cannot use a matrix workflow, because each runner will have to rebuild the disk, so tests will take forever.
> * It kinda pushes the GH runner storage to the limit.

We can configure another (paid) runner type with more CPU and storage. It's not going to make a dramatic difference (i.e. still no matrix), but it should help a bit.

yfyf:
> We can configure another (paid) runner type with more CPU and storage. It's not going to make a dramatic difference (i.e. still no matrix), but it should help a bit.

Yes, I wouldn't bother with that right now, but just something to keep in mind for the future.

The e2e test is failing because apparently there's some non-determinism in the kiosk persistence test when checking RAUC status, I can reproduce the failure locally, so not related to running it the GH workflow.

PR #189, opened 2024-10-17
Qt 6 reprise

knuton:
This reverts the Qt 6 revert (https://github.com/dividat/playos/pull/169) and makes the necessary changes to allow for persistent profiles, including using those created by a previous Qt 5 installation.

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/testing/integration/kiosk-proxy.nix
+++ b/testing/integration/kiosk-proxy.nix
@@ -1,0 +1,72 @@
+let
+  pkgs = import ../../pkgs { };
+  serverPort = 8080;
+  proxyPort = 8888;
+  kiosk = import ../../kiosk {
+    pkgs = pkgs;
+    system_name = "PlayOS";
+    system_version = "1.0.0";
+  };
+  toString = builtins.toString;
+in
+pkgs.nixosTest {
+  name = "proxy-test";
+
+  nodes = {
+    client = { config, ... }: {
+      imports = [
+        (pkgs.importFromNixos "tests/common/user-account.nix")
+        (pkgs.importFromNixos "tests/common/x11.nix")
+      ];
+
+      # Override is needed to enable in test VM, see connman tests:
+      # https://github.com/NixOS/nixpkgs/blob/1772251828be641110eb9a47ef530a1252ba211e/nixos/tests/connman.nix#L47-L52
+      services.connman.enable = pkgs.lib.mkOverride 0 true;
+
+      # We need a graphical environment and regular user for the kiosk browser
+      services.xserver = {
+        enable = true;
+      };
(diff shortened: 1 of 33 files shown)


knuton:
The e2e tests for kiosk persistence fail at 59e295c, but pass with the profile fix in 1753ea2.

fda922b should then additionally allow two-way compatibility with profiles created with a previous Qt 5 kiosk.

knuton:
Checked Qt5->Qt6 session persistence with https://dist.dividat.com/releases/playos/develop/2024.7.0-DEVELOP/playos-installer-2024.7.0-DEVELOP.iso as Qt 5 installation, which updates itself to https://dist.dividat.ch/releases/playos/develop/2024.10.0-SNAPSHOT/playos-installer-2024.10.0-SNAPSHOT.iso (viz, the associated bundle), built and released from fda922bc041ef2db6f6c37df1a0336b556eab56d.

After creating the session in the Qt5 system, and updating to the Qt6 system, the session is persisted. Switching back to the Qt5 system then after having booted into the Qt6 system, the session is still usable.

testing/integration/kiosk-proxy.nix
```diff
@@ -0,0 +1,72 @@
...
+      client.succeed(f"connmanctl config {service_name} proxy manual http://user:p4ssw0rd@theproxy:${toString proxyPort}")
+
+      kiosk_result = client.execute(
+        'su - alice -c "kiosk-browser http://thecloud:${toString serverPort} http://foo.xyz" 2>&1',
+        check_return=False,
+        check_output=True,
+        timeout=10
+      )
+
+      # Ideally here we would check if starting the kiosk resulted in a request
```
yfyf, line 57: Accidentally spotted this, but this is quite easily doable by e.g. setting up tinyproxy on the other VM and checking tinyproxy's logs: https://github.com/dividat/playos/blob/e2a4379fafd085c7ec9ad70df81c1e4e8a70ba9f/testing/integration/controller-proxy.nix#L45-L60

knuton: Ah yeah, thanks for the input.

The problem as I recall had not been inspecting the proxy, but getting the playos peer to resolve the names. I don't recall all details now, but think you may have found a solution in your recent tests, so it's time to look at this again.

knuton: Yes, your technique applies, I added back the full test based on that. :pray:

knuton:
@guyonvarch The e2e failure was unrelated and disappeared on re-run. May be a flaky test, but unrelated to this change as far as I can tell.

PR #191, opened 2024-10-21
Attempt to eliminate kiosk persistence e2e test flakiness

yfyf:
Two sources of flakiness:
- `rauc status` reporting `booted` as `null` and `rauc status` randomly crashing with a core dump. Always [on this line](https://github.com/yfyf/playos/blob/bc94691f30b99790b6879de2d090fba5be663855/testing/end-to-end/tests/application/kiosk-persistence.nix#L195). No idea why it happens, probably some uninitialized state, because it never crashes on the earlier calls? It might be related to the fact that the RAUC systemd service confiuration is incomplete, because it is missing [DBus activation specs](https://rauc.readthedocs.io/en/latest/integration.html#d-bus-integration).  Attempting to overcome this with a retry, adding the [necessary DBus service](https://github.com/ejoerns/rauc/blob/master/data/de.pengutronix.rauc.service.in) might be a good idea in general though.
- kiosk reporting that it has a `:` as the open page after being killed. Again no clue why it happens, but I am guessing we are attempting to interact with too early. Attempting to fix it by giving time to start before attempting to connect to the debug port.

Ran the test locally more than 10 times without failures. Before these changes would trigger both failures after ~5 runs.

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/testing/end-to-end/tests/application/kiosk-persistence.nix
+++ b/testing/end-to-end/tests/application/kiosk-persistence.nix
@@ -130,9 +130,15 @@ async def check_web_storages_after_restart(page, t):
             "TEST_VALUE cookie was not persisted"
         )
 
-def get_booted_slot():
+def get_booted_slot(retries=3):
     rauc_status = json.loads(playos.succeed("rauc status --output-format=json"))
-    return rauc_status['booted']
+    booted = rauc_status['booted']
+    # RAUC sometimes returns `null`, not sure why
+    if (booted is None) and retries > 0:
+        time.sleep(2)
+        return get_booted_slot(retries=retries-1)
+    else:
+        return booted
 
 # ===== Test scenario
 
--- a/testing/end-to-end/tests/application/kiosk-persistence.nix
+++ b/testing/end-to-end/tests/application/kiosk-persistence.nix
@@ -140,6 +140,11 @@ def get_booted_slot(retries=3):
     else:
         return booted
 
+def wait_for_dm_restart():
+    wait_for_logs(playos, "display-manager.service: Scheduled restart job")
+    playos.wait_for_unit("graphical-session.target")
+
+
 # ===== Test scenario
 
 aio = asyncio.Runner()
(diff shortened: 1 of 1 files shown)


knuton:
> * It might be related to the fact that the RAUC systemd service confiuration is incomplete, because it is missing [DBus activation specs](https://rauc.readthedocs.io/en/latest/integration.html#d-bus-integration).  Attempting to overcome this with a retry, adding the [necessary DBus service](https://github.com/ejoerns/rauc/blob/master/data/de.pengutronix.rauc.service.in) might be a good idea in general though.

Hm, you mean the `SystemdService` property? I understand that to be for scenarios where rauc should be started from dbus as needed, but not the only way to use rauc with dbus. Or are you referring to the [service type `dbus` as used here](https://github.com/ejoerns/rauc/blob/master/data/rauc.service.in)?

yfyf:
> Hm, you mean the SystemdService property? I understand that to be for scenarios where rauc should be started from dbus as needed, but not the only way to use rauc with dbus.

Yes and, true, it is not the only way. However, for some reason in e2e tests I have seen `rauc status` calls reporting that nothing provides the `de.pengutronix.rauc` bus, even though the tests were awaiting the start of `rauc.service`. Might be something broken in the `rauc.service` initialization.

knuton:
> However, for some reason in e2e tests I have seen `rauc status` calls reporting that nothing provides the `de.pengutronix.rauc` bus, even though the tests were awaiting the start of `rauc.service`. Might be something broken in the `rauc.service` initialization.

Quickly pushed https://github.com/dividat/playos/pull/192 to see whether we can redefine the service like this. Maybe this is going to make systemd aware that `rauc.service` is only to be considered up when the name is registered on the bus: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#id-1.10.6

Maybe it'll help prevent the kind of race condition we seem to hit in the test.

PR #194, opened 2024-10-23
Add end-to-end test report generation (and minor refactoring)

yfyf:
My attempt to produce a nice e2e test summary / report has turned into a ridiculous yak-shaving mission.

Some background:
- Spent hours Googling for simple CLI tools for language-agnostic automated testing of binaries, generating test reports, test report formats supported by GH, etc.
- At first was hoping to piggyback on [BATS](https://bats-core.readthedocs.io/) and it's JUnit output + [this GH Test Reporter](https://github.com/marketplace/actions/test-reporter). Issues: buggy (produces broken XML output), no option to suppress log output and just print a summary to console after running tests. Almost works, but not quite.
- I don't even want to get started on [TAP and the tooling](https://testanything.org/). What a mess.
- Tried generating a [CTRF report](https://ctrf.io/), but the generated GH summary is kinda ridiculous and there's no tooling for printing a summary to console.

So... decided to roll manually :shrug:
 
I am not very pleased with the outcome, but it's still better than nothing. If you have any ideas on how to do this in a neater way I am all ears.

Some points to note:
- `passthru.tests` (i.e. `nix-build -A test <..> default.nix`) never fails now, instead the exist status is stored in `$out/status` (see [build](https://github.com/yfyf/playos/blob/74442e178ba8fe4a180cb84bc2564087c90f7596/build#L132-L133)). I guess this could be avoided by `using --keep-failed` instead and getting the test run artifacts from the temporary build dir?
- Normal and interactive e2e test versions are stored in `./result` with naming based on their source filepath, i.e.:
    ```
    tree -l ./result/tests/ | sed 's/->.*$//'
    ./result/tests/
    ├── interactive
    │   ├── application
    │   │   └── kiosk-persistence
    │   └── base
    │       └── proxy-and-update
    └── tests
        ├── application
        │   └── kiosk-persistence
        └── base
            └── proxy-and-update
    ``` 

Example of a test summary printed to the console:
```
## End-to-end test result summary:
- base/proxy-and-update: FAIL ✗ (duration: 0:02:42)
- application/kiosk-persistence: OK ✓ (duration: 0:01:12)

Ran 2 tests, passed: 1, failed: 1
```

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,6 +81,9 @@ jobs:
     - name: Make magic-nix-cache read-only by removing post-build-hook
       run: sed -i '/post-build-hook = magic-nix-cache-build-hook/d' $HOME/.config/nix/nix.conf
     - run: ./build test-e2e
+    - name: Add summary
+      run: cat test-output/test-report.md >> "$GITHUB_STEP_SUMMARY"
+      if: always()
 
   ocaml-tests:
     runs-on: ubuntu-latest
--- a/build
+++ b/build
@@ -126,11 +126,11 @@ elif [ "$TARGET" == "test-e2e" ]; then
     --arg buildDisk false \
     --arg buildTest true
   "
-  echo "Building interactive e2e test runners."
+  echo "Building e2e test runners."
   (set -x; nix-build $test_flags)
   echo "Running e2e tests..."
-  (set -x; nix-build $test_flags -A tests --no-out-link)
-  echo "Done. All e2e tests passed."
+  (set -x; nix-build $test_flags -A tests -o test-output)
+  exit $(cat test-output/status)
 
 elif [ "$TARGET" == "all" ]; then
 
--- a/default.nix
+++ b/default.nix
@@ -156,10 +156,10 @@ with pkgs; stdenv.mkDerivation {
   ''
   # Tests
   + lib.optionalString buildTest ''
-    ln -s ${testComponents.tests.interactive-tests} $out/tests
+    mkdir -p $out/tests
+    ln -s ${testComponents.tests.interactive} $out/tests/interactive
+    ln -s ${testComponents.tests.tests} $out/tests/tests
   '';
 
-  passthru.tests = lib.optionalAttrs buildTest {
-    end-to-end = testComponents.tests.run-tests;
-  };
+  passthru.tests = testComponents.tests.run;
 }
--- a/testing/end-to-end/default.nix
+++ b/testing/end-to-end/default.nix
@@ -5,16 +5,85 @@ let
     overlayPath = "/tmp/playos-test-disk-overlay.qcow2";
     # fileFilter is recursive, so tests can in theory be in subfolders
     testFiles = fileset.fileFilter (file: file.hasExt "nix") ./tests;
-    testPackages = map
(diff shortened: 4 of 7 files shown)


yfyf:
GH test summary can be seen here:
https://github.com/yfyf/playos/actions/runs/11482660452#summary-31956262032

knuton:
> So... decided to roll manually 🤷

Right. I had a go at TAP/bats myself some time ago, and walked way from it. It always feels like it should be good, though. Maybe something hasn't clicked yet.

Help me understand the implications for our codebase. If you had done it via bats/TAP, would we have used a library that would have generated/help generate the right output? And then the tool would have transformed that to some summary? Or would we have still generated the right format of output ourselves, and only apply the tool to create the summary? Maybe the only extra tax is creating that markdown report manually?

Output in your failed test looks like a good thing! On `main` right now, it's really hard to see what failed at a glance.

yfyf:
> Help me understand the implications for our codebase. If you had done it via bats/TAP, would we have used a library that would have generated/help generate the right output? And then the tool would have transformed that to some summary? 

The BATS approach was:
- Wrap each e2e test (driver) `$fooTest` into a shell  script `${fooTest}.bats` containing ```@test "$fooTest" { $fooTest }```
- Link farm all the tests into some `$testDir`
- Run the tests with `bats --timing --recursive --report-formatter junit -o test-report.xml $testDir`

Issues with this:
1. No console output while tests are running, which sucks when running e2e tests because they take a while, might be hanging, etc. No flags to change this. No flags for changing this behaviour.
2. The generated `test-report.xml` is broken because BATS doesn't escape the log output (could be fixed by piping the output in the wrapper files).
3. If any test fails, it will dump all of the logs, no option to suppress this.
4. Final summary printed contains only total counts.

3 + 4 means that if anything fails, there's no good way to see an overview in the console, because individual test outcomes are interspersed by the failing test logs.

If it were only for the GH summary report, I would say that yeah, BATS + post-processing the XML is probably a better option. But now it just seems like a big indirection with questionable benefits.

yfyf:
Log formatting is now properly preserved in the summary page:
https://github.com/dividat/playos/actions/runs/11484431866?pr=194#summary-31962190652

knuton:
Rebased with `main`.

knuton (approved):
I think this is good to merge? I rebased and dropped the sample error commit.

>  `passthru.tests` (i.e. `nix-build -A test <..> default.nix`) never fails now, instead the exist status is stored in `$out/status` (see [build](https://github.com/yfyf/playos/blob/74442e178ba8fe4a180cb84bc2564087c90f7596/build#L132-L133)).

This seems relatively unproblematic for our current needs. We do run them in CI. I guess it might surprise any consumers of the top-level package, but to our knowledge the only "external" consumer is also us.

knuton:
> I am just thinking that it would be also nice if the CLI summary report was re-used for integration tests as well. We have 9 of them now and the current for-loop fail-on-first-err in `./bin/test` implementation is not ideal for local development. A summary table on Github would be helpful too.

:+1: 

> This could be a separate PR tho.

:+1:

PR #195, opened 2024-10-23
Use actions/checkout@v4

knuton:
v3 uses Node 16, which is EOL and causes warnings on GH Actions.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


yfyf (approved):
Nice, was itching to do the same, merging!

PR #198, opened 2024-10-30
Improve wifi error reporting and add tests

yfyf:
This was quite a Sisyphean task, but it does slightly improve the error reporting:
- Should now receive clear errors when attempting to connect to 802.1x networks
- Should now clearly indicate cases when passphrase is missing or AP demands additional inputs
- Messages now provide additional hints on what might be going wrong.

However, in most "weird" cases connman will respond with an `invalid-key` (for pretty much any wpa_supplicant error) or `connect-failed` (for other misc cases) without any details, which is not very helpful and this PR does not improve the situation. I noticed forked versions of connman in the wild which attempt to put more effort into handling different errors, but the vanilla version does not even use many of the error codes it defines (hence the commented out `manager_error` variants). 

Some extra notes:
- added a JSON version of the `/network` route and a JSON error handler for testing purposes. The error handler can in theory be removed (tests instead can rely on the raw HTML output), but the JSON output of `/network` is really needed for interpreting the output.
- introduced `yojson` because the PPX derivering libraries for Ezjson are kinda broken. Adds tech debt for ezjson deprecation and adds an extra serialization format alongside sexp.
- connman... kinda sucks?

TODOs:
- This deals with error message content, but we discussed it might make sense to also improve the error presentation. Separate PR for this maybe?
- What other type of APs / scenarios should we simulate / test? Doing something advanced requires intricate knowledge of wireless protocols and adds questionable benefits unless we see a way to improve connman's error reporting or want to dig into wpa_supplicant error reporting "out-of-band". I considered simulating noisy/lossy connections, but this would require setting up [wmediumd](https://github.com/bcopeland/wmediumd) and will most likely lead to "flaky" tests.
- No manual testing done. 

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/controller/bindings/connman/connman.ml
+++ b/controller/bindings/connman/connman.ml
@@ -174,6 +174,14 @@ struct
     | Online
   [@@deriving sexp, yojson]
 
+  type security =
+    | None
+    | WEP
+    | PSK
+    | IEEE8021x
+    | WPS
+  [@@deriving sexp, yojson]
+
   module IPv4 =
   struct
     type t = {
@@ -318,6 +326,7 @@ struct
   ; id : string
   ; name : string
   ; type' : Technology.type'
+  ; security: security list
   ; state : state
   ; strength : int option
   ; favorite : bool
@@ -341,6 +350,17 @@ struct
       | "disconnect" -> Some Disconnect
       | "online" -> Some Online
       | _ -> None
+    in
+    let security_of_string = function
+      | "none" -> Some None
+      | "psk" -> Some PSK
+      | "wps" -> Some WPS
+      | "ieee8021x" -> Some IEEE8021x
+      | _ -> None
+    in
+    let security_of_obus v =
+        let str_list = string_list_of_obus v in
+        List.filter_map security_of_string str_list
     in
     let strength_of_obus v =
       try
@@ -367,13 +387,13 @@ struct
         _ -> None
     in
     CCOption.(
-      pure (fun name type' state strength favorite autoconnect ipv4 ipv4_user_config ipv6 ethernet proxy nameservers ->
+      pure (fun name type' state strength favorite autoconnect ipv4 ipv4_user_config ipv6 ethernet proxy nameservers security ->
           { _proxy = OBus_proxy.make ~peer:(OBus_context.sender context) ~path:path
           ; _manager = manager
           ; id = path |> CCList.last 1 |> CCList.hd
           ; name ; type'; state; strength; favorite; autoconnect
           ; ipv4 = (if Option.is_some ipv4_user_config then ipv4_user_config else ipv4)
-          ; ipv6; ethernet; proxy; nameservers
+          ; ipv6; ethernet; proxy; nameservers; security
           })
       <*> (properties |> List.assoc_opt "Name" >>= string_of_obus)
       <*> (properties |> List.assoc_opt "Type" >>= string_of_obus >>= Technology.type_of_string)
@@ -386,7 +406,8 @@ struct
       <*> (properties |> List.assoc_opt "IPv6" >>= IPv6.of_obus |> pure)
       <*> (properties |> List.assoc_opt "Ethernet" >>= Ethernet.of_obus)
       <*> (properties |> List.assoc_opt "Proxy" >>= proxy_of_obus |> pure)
-      <*> (properties |> List.assoc_opt "Nameservers.Configuration" >|= string_list_of_obus))
+      <*> (properties |> List.assoc_opt "Nameservers.Configuration" >|= string_list_of_obus)
+      <*> (properties |> List.assoc_opt "Security" >|= security_of_obus))
 
   let is_connected t =
     match t.state with
@@ -454,7 +475,7 @@ struct
     in
 
     (* Store agent error in a local mutable variable *)
-    let agent_reported_error = ref None in
+    let agent_reported_error = ref Option.None in
     let on_agent_error msg = Lwt.return (agent_reported_error := Some msg) in
 
     (* Create and register an agent that will pass input to ConnMan *)
--- a/controller/bindings/connman/connman.ml
+++ b/controller/bindings/connman/connman.ml
@@ -86,56 +86,128 @@ struct
 
   type t = input OBus_object.t
 
+  type manager_error =
+      | InvalidKey
+      | ConnectFailed
+      | Blocked
+      (* see manager_error_of_string
+      | OutOfRange
+      | PinMissing
+      | DhcpFailed
+      | LoginFailed
+      | AuthFailed
+      *)
+      | Unknown of string
+
+  type agent_error =
+      | PassMissing
+      | MissingRequestedInputs of string list
+      | BrowserRequested
+      | UnexpectedFailure of exn
+
+  type reported_error =
+      | ManagerError of manager_error
+      | AgentError of agent_error
+
   let request_input input service (fields: (string * OBus_value.V.single) list) =
     let%lwt () = Logs_lwt.debug ~src:log_src
         (fun m -> m "input requested from agent for service %s"
             (String.concat "/" service)
         )
     in
-    match input with
-    | Passphrase p ->
-      (match List.assoc_opt "Passphrase" fields with
-       | Some _ ->
-         return [ "Passphrase", p |> OBus_value.C.(make_single basic_string)]
-       | None ->
-         let%lwt () = Logs_lwt.err ~src:log_src
-             (fun m -> m "Passphrase available as input but not being requested")
-         in
-         OBus_error.make "net.connman.Agent.Error.Canceled" "."
-         |> Lwt.fail
-      )
-    | None ->
-      let%lwt () = Logs_lwt.err ~src:log_src
-          (fun m -> m "input requested from agent but none available.")
-      in
-      OBus_error.make "net.connman.Agent.Error.Canceled" "No input available."
-      |> Lwt.fail
-
+    let pass_requested = List.assoc_opt "Passphrase" fields in
+    match (input, pass_requested) with
+        | (Passphrase p, Some _) ->
+            (* TODO: in theory manager can request passphrase _and_ other
+               fields, so this case is kind of incomplete? *)
+            return @@ Ok [ "Passphrase", p |> OBus_value.C.(make_single basic_string)]
+        | (None, Some _) ->
+          let%lwt () = Logs_lwt.err ~src:log_src (fun m -> m
+            "Passphrase requested from agent, but not available.")
+          in
+          return @@ Error PassMissing
+        | (_, None) ->
+            (* This case is triggered when manager is requesting
+               _some_ input(s) (so NOT open networks) and those inputs do not
+               involve a passphrase.
+            *)
+            let (expected_properties, _) = List.split fields in
+            let expected_properties_str =
+                String.concat ", " expected_properties
+            in
+            let%lwt () = Logs_lwt.err ~src:log_src (fun m -> m
+               "Manager is requesting additional missing input(s): %s"
+               expected_properties_str
+            )
+            in
+            return @@ Error (MissingRequestedInputs expected_properties)
 
   let request_browser input service url =
     let%lwt () = Logs_lwt.err ~src:log_src
         (fun m -> m "agent requested to open browser to url: %s" url)
     in
-    OBus_error.make "net.connman.Agent.Error.Canceled" "Can not open browser."
-    |> Lwt.fail
+    return @@ Error BrowserRequested
+
+  (* based on `connman_service_error` enum and it's stringified mapping in
+    error2string: https://git.kernel.org/pub/scm/network/connman/connman.git/tree/src/service.c?h=1.42#n324 *)
+  let manager_error_of_string = function
+    | "invalid-key" -> InvalidKey
+    | "connect-failed" -> ConnectFailed
+    (* although this sounds generic, `blocked` err is only produced by connman
+       when wpa_supplicant reports a wifi deauth with reason code = 6, which is
+       "Class 2 frame received from nonauthenticated station".
+       This seems like a protocol error? *)
+    | "blocked" -> Blocked
+(*
+    (* defined, but never produced in connman v1.42 *)
+    | "dhcp-failed" -> DhcpFailed
+    | "out-of-range" -> OutOfRange
+    | "pin-missing" -> PinMissing
+
+    (* these are VPN related *)
+    | "login-failed" -> LoginFailed
+    | "auth-failed" -> AuthFailed
+*)
+    | s -> Unknown s
+
+  let agent_error_msg = function
+      | PassMissing ->
+          "Passphrase is required for this access point"
+      | MissingRequestedInputs props ->
+          Printf.sprintf
+            "Unsupported protocol: additional inputs are needed: %s"
+            (String.concat ", " props)
+      | BrowserRequested ->
+          "Unsupported protocol: access point requires browser-based authentication"
+      | UnexpectedFailure exn ->
+          Printf.sprintf "Unexpected connection failure: %s" (Printexc.to_string exn)
 
   let interface on_error =
+    let wrap_req f = fun obj (x1, x2) ->
+        let%lwt caught_f = Lwt_result.catch
+            (fun () -> f (OBus_object.get obj) x1 x2) |>
+            (Lwt_result.map_error (fun e -> UnexpectedFailure e))
+        in
+        match (Result.join caught_f) with
+            | Ok a -> Lwt.return a
+            | Error err ->
+                let%lwt () = on_error (AgentError err) in
+                let obus_exn = OBus_error.make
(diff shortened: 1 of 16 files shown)


controller/bindings/connman/connman.ml
```diff
@@ -449,36 +546,75 @@ struct
...
+            service._proxy)
+    in
+    let%lwt _ = Lwt_result.catch destroy_agent in
+    match (obus_resp, !agent_reported_error) with
+        | (Ok _, _) ->
+            Lwt.return ()
+        | (Error _, Some (ManagerError InvalidKey)) ->
+            Lwt.fail_with "Passphrase is not valid or client is blocked. Please check the passphrase and then try to connect again."
+        | (Error _, Some (ManagerError ConnectFailed)) ->
+            Lwt.fail_with "Connection to network failed."
```
knuton, line 582: I know we don't know awfully much, but can we give any more hints here?

Is it fair to say "Connection to network failed for unknown reasons."? Even that seems less infuriating than "Connection to network failed.".

I think this is the case where it's worth saying "Please check signal strength and WiFi settings."?

controller/bindings/connman/connman.ml
```diff
@@ -86,56 +86,128 @@ struct
...
-      in
-      OBus_error.make "net.connman.Agent.Error.Canceled" "No input available."
-      |> Lwt.fail
-
+    let pass_requested = List.assoc_opt "Passphrase" fields in
+    match (input, pass_requested) with
+        | (Passphrase p, Some _) ->
+            (* TODO: in theory manager can request passphrase _and_ other
+               fields, so this case is kind of incomplete? *)
+            return @@ Ok [ "Passphrase", p |> OBus_value.C.(make_single basic_string)]
```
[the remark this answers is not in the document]
yfyf: > We could give up with an error when more than Passphrase is requested, but maybe that's too pessimistic

Looked at [connman's docs](https://git.kernel.org/pub/scm/network/connman/connman.git/tree/doc/agent-api.txt?h=1.42#n47) on `Agent.RequestInput`, and it's actually quite straightforward: connman flags each non-optional requested property as `Requirement: mandatory`, so we can check for those and error right away if there's anything mandatory beyond a passphrase. This is no longer "pessimistic" since connman will outright refuse to connect to if you fail to provide these inputs.  

From the docs: 

> This method gets called when trying to connect to
> a service and some extra input is required. For
> example a passphrase or the name of a hidden network.

Side-note, but If I understand correctly, the hidden network case is unsupported anyway, since it will simply not show up on the AP list.

> and memorize the list of requested inputs, so we can report it in case of subsequent failure

This would be a bit tricky to implement this, since we'd have to thread through a lot of state and it will make the code even more complex. I would just err to the user right away as you suggested.

testing/integration/controller-wifi.nix
```diff
@@ -0,0 +1,331 @@
...
+
+        # === dnsmasq + static IPs for wlan0*
+
+        # Setup networking + DHCP (dnsmasq) for the simulated wireless APs.
+        # Without this connman will wait ages for a (non-existant) DHCP
+        # offer, makin the tests unbearably slow.
+        #
+        # Note: this could probably be made to properly route traffic via the
+        # test VLANs or QEMU's vnet, but for now just returns some values in
+        # the 10.0.9.0/24 subnet
```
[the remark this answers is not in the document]
yfyf: Yes, all of these are easily doable and quite useful. I would suggest a separate PR for the `run-in-vm` improvements:
- incorporate the fake RAUC module to see the status output in the GUI
- add one functional wireless AP (with routing to the QEMU user network / host's gateway)
- maybe add one broken AP (can't connect to or can connect to but no internet)?

controller/bindings/connman/connman.ml
yfyf: Re-checked the source code, `connect-failed` is produced by connman in 3 cases:
- on association error (`CONNMAN_NETWORK_ERROR_ASSOCIATE_FAIL`) - does not seem to be used with wpa_supplicant, so irrelevant
- (network) configuration error (`CONNMAN_NETWORK_ERROR_CONFIGURE_FAIL`) - DHCP or IP configuration failed
- connection error (`CONNMAN_NETWORK_ERROR_CONNECT_FAIL`) - this seems to happen if e.g. connman cancels the connection process or gets disconnected from the AP, but I don't understand it fully - can be generalized as "some wpa_supplicant issue"

So in theory it can be anything, both wpa_supplicant errors (association/authentication issues) and network configuration problems. It sucks that connman doesn't propagate the more specific error to the outside world.

Rephrased in c16ca09, I think that's all we can do for now.

controller/bindings/connman/connman.ml
```diff
@@ -449,36 +546,75 @@ struct
...
+    let%lwt obus_resp = Lwt_result.catch (OBus_method.call
+            Connman_interfaces.Net_connman_Service.m_Connect
+            service._proxy)
+    in
+    let%lwt _ = Lwt_result.catch destroy_agent in
+    match (obus_resp, !agent_reported_error) with
+        | (Ok _, _) ->
+            Lwt.return ()
+        | (Error _, Some (ManagerError InvalidKey)) ->
+            Lwt.fail_with "Passphrase is not valid or client is blocked. Please check the passphrase and then try to connect again."
```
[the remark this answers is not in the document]
yfyf: Renamed it in 88fe08725e to the match the GUI naming.

controller/bindings/connman/connman.ml
```diff
@@ -82,60 +82,132 @@ struct
...
+      | InvalidKey
+      | ConnectFailed
+      | Blocked
+      (* see manager_error_of_string
+      | OutOfRange
+      | PinMissing
+      | DhcpFailed
+      | LoginFailed
+      | AuthFailed
+      *)
```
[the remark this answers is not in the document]
yfyf: Done in cd8603dd3d

controller/bindings/connman/connman.ml
knuton: Alright, thanks.

controller/bindings/connman/connman.ml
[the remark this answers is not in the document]
knuton: Great :+1: 

Looks like tests need to be updated to match it.

controller/bindings/connman/connman.ml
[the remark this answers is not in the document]
knuton: > connman flags each non-optional requested property as `Requirement: mandatory`, so we can check for those and error right away if there's anything mandatory beyond a passphrase. This is no longer "pessimistic" since connman will outright refuse to connect to if you fail to provide these inputs.

OK, sounds good

> If I understand correctly, the hidden network case is unsupported anyway, since it will simply not show up on the AP list.

That's right, we currently don't support them.

> This would be a bit tricky to implement this, since we'd have to thread through a lot of state and it will make the code even more complex. I would just err to the user right away as you suggested.

OK

controller/bindings/connman/connman.ml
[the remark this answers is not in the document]
yfyf: Fixed in 0de2bf406c

controller/bindings/connman/connman.ml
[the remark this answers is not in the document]
yfyf: Implemented in 48c66c68fa0

testing/integration/controller-wifi.nix
```diff
@@ -0,0 +1,331 @@
...
+              # connman hangs when connecting to this AP, nothing in the logs
+              # neither from connman nor from wpa_supplicant ¯\_(ツ)_/¯
+              #wlan0-2 = {
+              #  ssid = "test-ap-wpa2";
+              #  authentication = {
+              #    mode = "wpa2-sha256";
+              #    wpaPassword = "reproducibility";
+              #  };
+              #  bssid = "02:00:00:00:00:02";
+              #};
```
knuton, line 138: I would be curious whether this issue persists with connman 1.43, because it sounds exactly like the behavior I saw with the paid network in my area (though that was unauthenticated + captive portal, so not an obvious match).

knuton:
> * added a JSON version of the `/network` route and a JSON error handler for testing purposes. The error handler can in theory be removed (tests instead can rely on the raw HTML output), but the JSON output of `/network` is really needed for interpreting the output.

Overhead is relatively low, so it's probably OK. Another option might be to include more information than strictly required in the HTML markup. I think the Beautiful Soup lib in Python makes pulling info out relatively convenient.

What data exactly do you need that goes beyond the current HTML?

> * introduced `yojson` because the PPX derivering libraries for Ezjson are kinda broken. Adds tech debt for ezjson deprecation and adds an extra serialization format alongside sexp.

Can you explain the issues you found with ezjson?

> * This deals with error message content, but we discussed it might make sense to also improve the error presentation. Separate PR for this maybe?

Yes, separate PR, in which:

- Suppress password input for open networks.
- Add note to networks with unsupported security type, or even hide connect button?
- Stay in "Network" page and extract known failures instead of letting HTTP error handler deal with connection errors.

> * What other type of APs / scenarios should we simulate / test? Doing something advanced requires intricate knowledge of wireless protocols and adds questionable benefits unless we see a way to improve connman's error reporting or want to dig into wpa_supplicant error reporting "out-of-band". I considered simulating noisy/lossy connections, but this would require setting up [wmediumd](https://github.com/bcopeland/wmediumd) and will most likely lead to "flaky" tests.

I agree, would focus on presenting what we have properly before going into advanced failure simulation.

> * No manual testing done.

In my manual testing I can confirm your sentiment that connman deals with errors poorly, meaning we get very undiscriminating error feedback. The situation is at least a lot better than before, though. Errors seem human and a little more helpful.

yfyf:
> Overhead is relatively low, so it's probably OK. Another option might be to include more information than strictly required in the HTML markup. I think the Beautiful Soup lib in Python makes pulling info out relatively convenient.
> 
> What data exactly do you need that goes beyond the current HTML?

It's not that it goes "beyond", but that in order to get the structural data the tests would have to do a lot of gymnastics (e.g. to get connman service id of each AP name, it would have to parse HTML +  [regex the HTML a href link](https://github.com/dividat/playos/blob/48c66c68fa00ffb497aec0645c01d10e1fb143f0/controller/server/view/network_list_page.ml#L19)). I also thought it could be useful in the future for other types of automated and/or manual testing. 

I guess an alternative is to add id's + data attributes to the HTML elements and extract those, though I don't find it particularly elegant (the presentation layer suddenly becomes the data layer?)

> > * introduced `yojson` because the PPX derivering libraries for Ezjson are kinda broken. Adds tech debt for ezjson deprecation and adds an extra serialization format alongside sexp.
> 
> Can you explain the issues you found with ezjson?

I was attempting to use it via `ppx_deriving_yaml`, which is comptible with ezjson, but doesn't support variant types. I checked again and there's an alternative of using [ppx_protocol_conv](https://github.com/andersfugmann/ppx_protocol_conv), I can try that?

> Yes, separate PR, in which:
> 
> * Suppress password input for open networks.
> * Add note to networks with unsupported security type, or even hide connect button?
> * Stay in "Network" page and extract known failures instead of letting HTTP error handler deal with connection errors.

:+1:

testing/integration/controller-wifi.nix
knuton: I just re-ran the tests with a nixpkgs bump merged in, and it completes when this is commented out, but fails when this AP is included.

Pity!

yfyf:
@knuton replaced `yojson` and `ppx_yojson_conv` with `ppx_protocol_conv`(`_jsonm`)

pkgs/ocaml-modules/ppx_protocol_conv/default.nix
```diff
@@ -0,0 +1,43 @@
...
+    owner = "andersfugmann";
+    repo = "ppx_protocol_conv";
+    rev = "9a1c4f450cf4fc797e09296d736e5c21f6eb7c4a";
+    sha256 = "sha256-3WlpuuRJfZA6UaeuJRuYI2Z9J7Yng+/J09BDmETwy30=";
+  };
+
+  useDune2 = true;
+
+  #nativeBuildInputs = [ ];
+  propagatedBuildInputs = [
```
knuton, line 28: Is `buildInputs` what you are looking for?

testing/integration/controller-wifi.nix
```diff
@@ -0,0 +1,331 @@
...
+              #wlan0-4 = {
+              #  ssid = "bad-ap-blocked";
+              #  bssid = "02:00:00:00:00:04";
+              #  authentication = {
+              #    mode = "wpa3-sae";
+              #    saePasswords = [ { password = "reproducibility"; } ];
+              #  };
+              #  macAcl = "allow"; # allows only listed clients
+              #  macAllow = [ ] ; # deny everything
+              #};
```
knuton, line 162: This seems like good info to capture, but the commented out code makes its status unclear to me.

Could this even be a test that runs, but we explain the unexpected message in the test code?

testing/integration/controller-wifi.nix
knuton: As long as we can not resolve this, shall we just more concisely document that mode `wpa2-sha256` seems to have issues?

A more drastic (and possibly correct) measure would be to deny connection requests to such networks and make them officially unsupported. They seem to be rare, with generally low support, and WPA3 is an alternative nowadays. But I guess we can not (easily) detect this from connman?

controller/bindings/connman/connman.ml
```diff
@@ -82,60 +89,135 @@ struct
...
+            (* This case is triggered when manager is requesting
+               _some_ input(s) (so NOT open networks) and those inputs
+               involve something additional to (or other than) a passphrase.
+            *)
+            let expected_properties_str =
+                String.concat ", " mandatory_inputs
+            in
+            let%lwt () = Logs_lwt.err ~src:log_src (fun m -> m
+               "Manager is requesting additional missing input(s): %s"
+               expected_properties_str
```
knuton, line 150: Nitpick: This would seem to imply that the password is missing, even when it has been entered. It's also not so clear to a potential users who the "Manager" is.

Rephrase to state that the _network_ is requesting missing inputs, and give the full list of inputs? (Alternative: Filter out passphrase from list if one was given.)

`"This network requires additional inputs for authentication. The network requests the following: %s"`

testing/integration/controller-wifi.nix
yfyf: It's quite an esoteric case, not really worthy of a test - I have simply removed it in 4b38813

pkgs/ocaml-modules/ppx_protocol_conv_jsonm/default.nix
```diff
@@ -0,0 +1,41 @@
...
+  src = fetchFromGitHub {
+    owner = "andersfugmann";
+    repo = "ppx_protocol_conv";
+    rev = "9a1c4f450cf4fc797e09296d736e5c21f6eb7c4a";
+    sha256 = "sha256-3WlpuuRJfZA6UaeuJRuYI2Z9J7Yng+/J09BDmETwy30=";
+  };
+
+  useDune2 = true;
+
+  #nativeBuildInputs = [ ];
```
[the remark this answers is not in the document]
yfyf: 3dde5d11b

pkgs/ocaml-modules/ppx_protocol_conv/default.nix
yfyf: 3dde5d11b

knuton: So using only `buildInputs` doesn't work?

testing/integration/controller-wifi.nix
knuton: OK, how about adding some context then that helps re-evaluate in the future?

Maybe just prefix the explanation with `As of connman 1.43, it hangs ...`?

controller/bindings/connman/connman.ml
knuton: Of course, I was in read-string-translations mode and completely ignored the surrounding lines ...

pkgs/ocaml-modules/ppx_protocol_conv/default.nix
yfyf: `buildInputs` works, but all the other packages in `pkgs/ocaml-modules` seem to favor `propagatedBuildInputs`, so I assumed that's the "right" target? Should I change the other modules as well then?

testing/integration/controller-wifi.nix
yfyf: Done in 4d62603

pkgs/ocaml-modules/ppx_protocol_conv/default.nix
knuton: Hmm ... you are right, also in nixpkgs most OCaml modules are packaged this way.

https://nixos.org/manual/nixpkgs/stable/#sec-language-ocaml-packaging does not shed too much wisdom on it.

I guess propagated is needed for transitive dependencies as the build only really happens at the root of the tree.

PR #203, opened 2024-11-19
Wifi GUI improvements

yfyf:
1. No passphrase prompt for open wifi APs
2. Prevent users from connecting to wifi APs with unsupported protocols

Tested manually by copying the setup from `testing/integration/controller-wifi.nix` in ./build vm.

## Checklist

-   [ ] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/server/view/network_details_page.ml
+++ b/controller/server/view/network_details_page.ml
@@ -77,24 +77,29 @@ let proxy_form proxy =
         ]
     ]
 
+let maybe_elem cond elem = if cond then Some elem else None
+
 let not_connected_form service =
+  let requires_passphrase =  service.security <> [ None ] in
   form
       ~a:[ a_action ("/network/" ^ service.id ^ "/connect")
       ; a_method `Post
       ; Unsafe.string_attrib "is" "disable-after-submit"
       ]
-      [ label
-          ~a:[ a_class [ "d-Label" ] ]
-          [ txt "Passphrase" 
-          ; input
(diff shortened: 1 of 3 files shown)


knuton:
> Tested manually by copying the setup from `testing/integration/controller-wifi.nix` in ./build vm.

Can we include this permanently?

controller/server/view/network_details_page.ml
```diff
@@ -225,8 +230,23 @@ let connected_form service =
...
+            (fun s -> Sexplib.Sexp.to_string (sexp_of_security s))
+            service.security))
+    ]
+
 let html service =
   let is_service_connected = Connman.Service.is_connected service in
+  let is_service_supported =
+      List.exists (fun e -> List.mem e service.security) [ PSK; WEP; None ]
+    ||
+      (service.security = []) (* wired connections have this *)
```
yfyf, line 248: Strictly speaking this extra case is not needed, because currently `is_service_supported` is only relevant for wifi connections, but I added it to make the code more future proof in case this changes somehow.

yfyf:
@knuton added a `testing-wifi` module in 4b50358, check it out.

knuton (approved):
Great :sparkles:

controller/server/view/network_details_page.ml
knuton: Seems reasonable :+1:

PR #204, opened 2024-11-19
Rename "passphrase" to "password" in GUI

knuton:
I had been under the impression we already used "password". We didn't, but we should.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


yfyf (approved):
Good catch. It is easy to get confused because the code / interfaces mostly use "passphrase"!

PR #205, opened 2024-11-20
Enable read-only usage of magic-nix-cache in e2e tests

yfyf:
Contrary to the documentation, turns out `use-gha-cache: false` completely disabled the GHA, i.e. everything.

Since we don't use `diff-store: true`, by removing the post-build-hook this makes the magic nix cache think that nothing was built (and hence there's nothing new to cache).

Verified nothing was added by [inspecting the Action Cache](https://github.com/dividat/playos/actions/caches?query=sort%3Asize-desc).


knuton (approved):

knuton:
> What has been will be again, what has been done will be done again; there is nothing new under the sun.

PR #206, opened 2024-11-21
Patch ppx_protocol_conv_jsonm to not have a dependency on ppxlib

yfyf:
See the following commit upstream for a similar change to the (yo)json driver:
https://github.com/andersfugmann/ppx_protocol_conv/commit/1be9d66f8f1ef60089d8c8f10b7135cc83fea40b

This avoids pulling in all of ppxlib (which is a good thing in general) and in particular does not pollute the global namespace with `compiler-lib` modules (which fixes the issues I was having with rebasing #194 onto `main`).

Will submit a patch upstream as well.

Side-note: it might make be useful at some point to review the current dependencies in general, because e.g. we link to `Base`, but the way it is used is redundant.

### DIFF
--- a/pkgs/ocaml-modules/ppx_protocol_conv_jsonm/default.nix
+++ b/pkgs/ocaml-modules/ppx_protocol_conv_jsonm/default.nix
@@ -23,6 +23,10 @@ buildDunePackage rec {
 
   useDune2 = true;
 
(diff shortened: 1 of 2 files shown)


knuton (approved):
Hallelujah!

PR #209, opened 2024-11-25
Split Update state into version info, system status and process state

yfyf:
This refactors Update state into 3 parts:
- `process_state`, which indicates what action the Update process is currently performing (checking, downloading, installing or sleeping)
- `system_status`, which reflects what the Update process thinks is the current system state (up-to-date, needs-update, etc)
- `version_info`, which reflects the current slot and latest upstream versions.

The reason for doing this is to decompose the process transitions from the (determined) system state, as consumed by other components via the `React.Signal`. For example, this allows to observe the last `system_status` in the GUI while the Update process is performing other tasks.

Note that only the `process_state` is an input state (meaning that the behaviour of Update process depends on it), whereas the `system_status` and `version_info` are "output" states.

In theory, the basic FSM step could be refactored into something like:

        val run_step: process_state
            -> (process_state * (system_status option) * (version_info option)) Lwt.t

where the recursive loop would now `run_rec` maintain the `state` and update it's fields if the output `option`s are set. However, this splits the logic over two functions rather than one and makes it more complicated to interpret state machine.

See inline comments for explanations on some of the changes.

## Checklist

-   [ ] Changelog updated
-   [X] Code documented
-   [ ] User manual updated

### DIFF
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -424,12 +424,12 @@ module StatusGui = struct
     ) in
     reboot ()
 
-  let get_status ~health_s ~update_s ~rauc =
+  let get_status ~health_s ~(update_s:Update.state React.signal) ~rauc =
         let health_state = health_s |> Lwt_react.S.value in
         let update_state = update_s |> Lwt_react.S.value in
         let%lwt booted_slot = Rauc.get_booted_slot rauc in
         let%lwt rauc =
-          match update_state with
+          match update_state.process_state with
           (* RAUC status is not meaningful while installing
              https://github.com/rauc/rauc/issues/416
           *)
--- a/controller/tests/server/update/outcome.ml
+++ b/controller/tests/server/update/outcome.ml
@@ -35,11 +35,11 @@ let state_matches_expected_outcome state outcome =
     | (DoNothingOrProduceWarning, UpToDate, Sleeping _) -> true
     | (DoNothingOrProduceWarning, OutOfDateVersionSelected, Sleeping _) -> true
     | (DoNothingOrProduceWarning, RebootRequired, Sleeping _) -> true
-    | (DoNothingOrProduceWarning, ReinstallRequired, Sleeping _) ->true
+    | (DoNothingOrProduceWarning, ReinstallRequired, Sleeping _) -> true
     (* should not _directly_ return to GettingVersionInfo state *) (*TODO: redundant now *)
     | (DoNothingOrProduceWarning, _, Update.GettingVersionInfo) -> false
     (* all the other state combos are treated as errors *)
-    | (DoNothingOrProduceWarning, _, _) ->false
+    | (DoNothingOrProduceWarning, _, _) -> false
 
 (** Tests if the input UpdateService run with the given [Helpers.system_slot_spec]
     [case] scenario produces the expected outcome state (defined by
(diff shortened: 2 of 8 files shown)


controller/server/gui.ml
```diff
@@ -424,12 +424,12 @@ module StatusGui = struct
     ) in
     reboot ()
 
-  let get_status ~health_s ~update_s ~rauc =
+  let get_status ~health_s ~(update_s:Update.state React.signal) ~rauc =
```
yfyf, line 427: ocamlc's type inference is failing here for some reason

controller/tests/server/update/outcome.ml
```diff
@@ -25,20 +25,21 @@ let slot_spec_to_outcome ({booted_slot; primary_slot; input_versions} : Helpers.
...
-        | (DoNothingOrProduceWarning, Update.OutOfDateVersionSelected) ->  true
-        | (DoNothingOrProduceWarning, Update.RebootRequired) ->            true
-        | (DoNothingOrProduceWarning, Update.ReinstallRequired) ->         true
-        (* should not _directly_ return to GettingVersionInfo state *)
-        | (DoNothingOrProduceWarning, Update.GettingVersionInfo) ->        false
-        (* all the other states are part of the installation process
-           and are treated as errors *)
-        | (DoNothingOrProduceWarning, _) ->                         false
+    let open Update in
+    match (outcome, state.system_status, state.process_state) with
```
yfyf, line 29: Note: the spec here is made stricter by verifying that the returned `version_info` is updated and that the process is in the expected `process_state`, but otherwise is not changed.

controller/tests/server/update/update_tests.ml
```diff
@@ -219,9 +244,10 @@ let () =
            [
              Alcotest_lwt.test_case "Delete downloaded bundle on install error"
              `Quick (fun _ () -> Scenario.run delete_downloaded_bundle_on_err);
-
-             Alcotest_lwt.test_case "Sleep for a duration after error or check"
```
yfyf, line 223: Replacing the sleep test (which was checking side-effects in a rather hacky way) with two different tests that check that Update transitions into `Sleeping` state after errors, because the "error" states actually no longer exist as process states.

knuton (commented):
This seems great!

I have only a nitpick and a sanity check I'll run tomorrow.

controller/tests/server/update/outcome.ml
```diff
@@ -25,20 +25,21 @@ let slot_spec_to_outcome ({booted_slot; primary_slot; input_versions} : Helpers.
...
+            (Semver.to_string v1) = v2 &&
+            Option.fold ~none:false ~some:(fun v -> v.latest = v1) state.version_info
+    | (InstallVsn _, _, _) -> false
+    | (DoNothingOrProduceWarning, UpdateError (ErrorGettingVersionInfo _), Sleeping _) -> true
+    | (DoNothingOrProduceWarning, UpToDate, Sleeping _) -> true
+    | (DoNothingOrProduceWarning, OutOfDateVersionSelected, Sleeping _) -> true
+    | (DoNothingOrProduceWarning, RebootRequired, Sleeping _) -> true
+    | (DoNothingOrProduceWarning, ReinstallRequired, Sleeping _) ->true
+    (* should not _directly_ return to GettingVersionInfo state *) (*TODO: redundant now *)
+    | (DoNothingOrProduceWarning, _, Update.GettingVersionInfo) -> false
```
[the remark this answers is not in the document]
yfyf: It should be merged, yes. It was redundant before and it remains redundant now. I _think_ I added the TODO, because now all the expected process states are made explicit, so it is "even more" redundant. Originally I did add it for documentation purposes, yes.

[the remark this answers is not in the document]
yfyf: Removed in 1fc88d5

knuton (approved):
Nice

At first I thought we would want to show only part of the update state in the GUI, but in the end I think it's useful to show all. At least system status and version info are both of general interest, and there would be times when process state is also insightful.

It's another very technical looking output field with the sexp serialization, but the explanations and action buttons for the various states bring the crucial information forward in a more accessible manner.

Formatting sexp structures in a less intimidating way in the GUI might be a job more general than this PR.

PR #184, opened 2024-10-07
Bump nixpkgs to 24.11

knuton:
## Tested

- [x] e2e
- [x] Release validation
- [x] VM
- [x] Live system
- [x] Installed system 
  - [x] RFID reader
  - [x] WebGL
  - [x] Senso connection
  - [x] WiFi
  - [x] Audio output

## To do

- [x] Audio output ~~reverted to internal speaker~~ not working on ASUS:
    - [x] opus codec not working with nixpkgs 24.11(specifically, after the upgrade to Qt 6.8.0 in NixOS/nixpkgs@d4f7b842c)
    - [x] audio device (card) ordering changed and `set-card-profile 0 ...` references the external USB card (`usb-MEMS_TECH_ad03-00`) - ~~should we blacklist `"usb"` in `module-switch-on-connect`?~~ blacklisting is not sufficient, does not affect ordering 
    - [x] in synthetic QEMU tests, since 24.11 the default sink is muted (might be due to missing hdmi profile and a non-issue, but it has changed since 24.05)
- [x] Failing integration test, maybe issue with test driver ([currently disabled](https://github.com/dividat/playos/pull/184/commits/5ff87c3c75546f4daa5af90e5d463797a5eeb1ac))
- [x] Nixpkgs can be bumped to latest released commit of 24.11 again
- [x] Power button special casing also seems broken, short press on built-in power button no longer causes poweroff: https://github.com/dividat/playos/blob/main/application/power-management/default.nix
- [x] Python syntax warning `/home/emerij/Development/playos/kiosk/kiosk_browser/browser_widget.py:148: SyntaxWarning: invalid escape sequence '\('` in `pattern = re.compile('(Mozilla/5.0) \(([^\)]*)\)(.*)')` (should probably just be `r'(Mozilla/5.0) \(([^\)]*)\)(.*)'`)
- [x] e2e tests:
   - [x] run out of space on GH Actions runners - verify?
   - [x] fix application/kiosk-persistence test
- [x] Go straight to Nixpkgs 24.11 ([release happening in late Nov](https://discourse.nixos.org/t/lets-have-a-great-24-11-release-cycle/51549))?
  - This brings connman 1.43, with fix of hanging connman with weird captive portal networks.
  - Otherwise could backport connman 1.43.
- [x] More warnings to squash with 24.11
- [x] Fix `./testing/manual/kiosk-dual-screen.nix`
- [x] Check why `systemImage` / test disk size has increased
[...]

PR #213, opened 2024-12-09
Cherry-picked CI additions

yfyf:
Cherry-picked these from #184, because they are needed in general and because I want to use the common CI step abstraction in #200


yfyf:
Whoops, pushed without testing. Amended and forced pushed, should be good now.

yfyf:
> Running `testing/run` works successfully on `main`, but on this branch, I get the following error:
> ```
> <..>
> [TestCase] kiosk-browser uses configured proxy... Fail!
> ```

Can you retry `nix-build -A driver testing/integration/kiosk-proxy.nix && ./result/bin/nixos-test-driver` a few times? This test is flaky, I have had it fail locally on different branches, so I think this is unrelated. It passes for me, I cannot trigger the failure currently. 

> Running `testing/run`, with everything commented except for manual tests, the script gets blocked:
>...
> Any idea?

Hmmm, what operating system are you using? Might be something failing due to `virtio-vga`? Are you able to run `./build vm` with the patch below? 

 ```
diff --git a/testing/run-in-vm/run-in-vm.py b/testing/run-in-vm/run-in-vm.py
index 4f70333..b2a5d83 100755
--- a/testing/run-in-vm/run-in-vm.py
+++ b/testing/run-in-vm/run-in-vm.py
@@ -64,7 +64,8 @@ def run_vm(system, qemu_opts, kernel_arguments):
             '-initrd', initrd,
             '--virtfs', virtfs_opts,
             '-append', kernel_arguments,
-            '-vga', 'virtio', # Display screen as big as possible
+            #'-vga', 'virtio', # Display screen as big as possible
+            '-device', 'virtio-vga,max_outputs=2',
             # Unused shell. This is used by the "backdoor" in <nixos/modules/test-instrumentation.nix>.
             # TODO: Set up a Unix Socket from Python and connect to shell
             '-chardev', 'socket,id=shell,path={}/not-working-shell,server=on,wait=off'.format(backdoor_dir),
```

PR #224, opened 2025-02-04
Support for multiple 'Power Button's

knuton:
The original special casing of 'Power Button' devices stops working (on the ASUS device) after upgrading to nixpkgs 24.11 (#184), because the system has more than one input devices with this name.

I added a basic integration test for the basic feature: Ignoring key presses from keyboards, but acting on presses from "Power Button" device.

I tried to extend the test for the multiple device scenario as well, unfortunately I ran into a seeming limitations of QEMU: It doesn't seem to be possible to rename either `usb-kbd` or `virtio-keyboard` in QEMU, so the test does not capture that multiple devices named "Power Button" are handled properly.

I also couldn't figure out how to inject long key presses.

It might be possible to use something like xdotool or evemu for complete testing.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/application/power-management/power-button-shutdown.py
+++ b/application/power-management/power-button-shutdown.py
@@ -2,22 +2,58 @@
 
 import evdev
 import logging
-import os
+import subprocess
 import sys
+import threading
 
 logging.basicConfig(level=logging.INFO)
 
-devices = [evdev.InputDevice(path) for path in evdev.list_devices()]
-device_names = ', '.join([d.name for d in devices])
-logging.info(f'Found devices: {device_names}')
+def handle_device(device):
+    """Handle power key presses for the given input device by invoking poweroff."""
 
-power_off_device = next((d for d in devices if d.name == 'Power Button'), None)
-if power_off_device is None:
-  logging.error(f'Power Button device not found')
-  sys.exit(1)
+    logging.info(f'Listening to Power Button on device {device.path}')
+    try:
+        for event in device.read_loop():
+            if event.type == evdev.ecodes.EV_KEY and event.code == evdev.ecodes.KEY_POWER and event.value:
+                logging.info(f'KEY_POWER detected on {device.path}, shutting down')
+                subprocess.run(['systemctl', 'start', 'poweroff.target'], check=True)
+    except Exception as e:
+        logging.error(f'Error handling {device.path}: {e}')
 
-logging.info(f'Listening to Power Button on device {power_off_device.path}')
-for event in power_off_device.read_loop():
-  if event.type == evdev.ecodes.EV_KEY and evdev.ecodes.KEY[event.code] == 'KEY_POWER' and event.value:
-    logging.info('KEY_POWER detected on Power Button device, shutting down')
-    os.system('systemctl start poweroff.target')
+def get_power_button_devices():
+    """Get all input devices which identify as 'Power Button' and have a power key."""
+
+    devices = [evdev.InputDevice(path) for path in evdev.list_devices()]
+
+    power_button_devices = []
+    for device in devices:
+        try:
+            capabilities = device.capabilities()
+            if device.name == 'Power Button' and evdev.ecodes.EV_KEY in capabilities and evdev.ecodes.KEY_POWER in capabilities[evdev.ecodes.EV_KEY]:
+                power_button_devices.append(device)
+        except Exception as e:
+            logging.warning(f'Failed to inspect {device.path}: {e}')
+
+    return power_button_devices
+
+# Identify Power Button devices
+power_button_devices = get_power_button_devices()
+if not power_button_devices:
+    logging.error('No Power Button devices found')
+    sys.exit(1)
+logging.info(f'Found {len(power_button_devices)} Power Button device(s)')
+
+# Start a thread with a handler for each identified device
+handler_threads = []
+for device in power_button_devices:
+    thread = threading.Thread(target=handle_device, args=(device,))
+    thread.start()
+    handler_threads.append(thread)
+
+# Exit gracefully if a handler executes
+try:
+    for thread in handler_threads:
+        thread.join()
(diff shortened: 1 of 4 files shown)


knuton:
For the record, I looked into options for renaming the QEMU keyboards as follows:

```
> qemu-system-x86_64 -device usb-kbd,help
usb-kbd options:
  attached=<bool>
  display=<str>
  msos-desc=<bool>       - on/off (default: true)
  pcap=<str>
  port=<str>
  serial=<str>
  usb_version=<uint32>   -  (default: 2)

> qemu-system-x86_64 -device virtio-keyboard,help
virtio-keyboard-pci options:
  acpi-index=<uint32>    -  (default: 0)
  addr=<int32>           - Slot and optional function number, example: 06.0 or 06 (default: -1)
  aer=<bool>             - on/off (default: false)
  any_layout=<bool>      - on/off (default: true)
  ats=<bool>             - on/off (default: false)
  disable-legacy=<OnOffAuto> - on/off/auto (default: "auto")
  disable-modern=<bool>  -  (default: false)
  display=<str>
  event_idx=<bool>       - on/off (default: true)
  failover_pair_id=<str>
  head=<uint32>          -  (default: 0)
  indirect_desc=<bool>   - on/off (default: true)
  iommu_platform=<bool>  - on/off (default: false)
  migrate-extra=<bool>   - on/off (default: true)
  modern-pio-notify=<bool> - on/off (default: false)
  multifunction=<bool>   - on/off (default: false)
  notify_on_empty=<bool> - on/off (default: true)
  packed=<bool>          - on/off (default: false)
  page-per-vq=<bool>     - on/off (default: false)
  queue_reset=<bool>     - on/off (default: true)
  rombar=<uint32>        -  (default: 1)
  romfile=<str>
  romsize=<uint32>       -  (default: 4294967295)
  serial=<str>
  use-disabled-flag=<bool> -  (default: true)
  use-started=<bool>     -  (default: true)
  vectors=<uint32>       -  (default: 2)
  virtio-backend=<child<virtio-keyboard-device>>
  virtio-pci-bus-master-bug-migration=<bool> - on/off (default: false)
  x-ats-page-aligned=<bool> - on/off (default: true)
  x-disable-legacy-check=<bool> -  (default: false)
  x-disable-pcie=<bool>  - on/off (default: false)
  x-ignore-backend-features=<bool> -  (default: false)
  x-pcie-ari-nextfn-1=<bool> - on/off (default: false)
  x-pcie-deverr-init=<bool> - on/off (default: true)
  x-pcie-err-unc-mask=<bool> - on/off (default: true)
  x-pcie-extcap-init=<bool> - on/off (default: true)
  x-pcie-flr-init=<bool> - on/off (default: true)
  x-pcie-lnkctl-init=<bool> - on/off (default: true)
  x-pcie-lnksta-dllla=<bool> - on/off (default: true)
  x-pcie-pm-init=<bool>  - on/off (default: true)
```

There doesn't seem to be a suitable option.

application/power-management/power-button-shutdown.py
```diff
@@ -2,22 +2,58 @@
...
+    """Get all input devices which identify as 'Power Button' and have a power key."""
+
+    devices = [evdev.InputDevice(path) for path in evdev.list_devices()]
+
+    power_button_devices = []
+    for device in devices:
+        try:
+            capabilities = device.capabilities()
+            if device.name == 'Power Button' and evdev.ecodes.EV_KEY in capabilities and evdev.ecodes.KEY_POWER in capabilities[evdev.ecodes.EV_KEY]:
+                power_button_devices.append(device)
```
yfyf, line 33: I find something like this a bit more readable (untested):

```suggestion
if device.name == 'Power Button':
    ev_capabilities = device.capabilities().get(evdev.ecodes.EV_KEY, [])
    if evdev.ecodes.KEY_POWER in ev_capabilities:
        power_button_devices.append(device)
```

yfyf:
> For the record, I looked into options for renaming the QEMU keyboards as follows:

Could this be releavant? https://stackoverflow.com/a/70146528

I tried running

    ./result/bin/run-in-vm -q -enable-kvm -m 2048 -device virtio-keyboard-pci,display=seat1 -device virtio-keyboard-pci,display=seat2

and can see multiple keyboards with lspci:

```
[root@playos-test:~]# lspci
<...>
00:07.0 Keyboard controller: Red Hat, Inc. Virtio 1.0 input (rev 01)
00:08.0 Keyboard controller: Red Hat, Inc. Virtio 1.0 input (rev 01)
```

> I tried to extend the test for the multiple device scenario as well, unfortunately I ran into a seeming limitations of QEMU: It doesn't seem to be possible to rename either usb-kbd or virtio-keyboard in QEMU

Hmm, could this be solved by handling the naming using udev rules? Although maybe it's not worth investing time, I don't think this needs to be tested that "deeply".

Overall, I don't really understand why the whole thing can't be handled as a udev rule, this seems a bit complicated (a threaded application, etc).

yfyf:
Would a correct `systemd-logind` configuration not be sufficient here? i.e. `HandlePowerKey=poweroff`?

application/power-management/power-button-shutdown.py
knuton: Nice

fb777c6

application/power-management/power-button-shutdown.py
```diff
@@ -2,22 +2,58 @@
...
+        except Exception as e:
+            logging.warning(f'Failed to inspect {device.path}: {e}')
+
+    return power_button_devices
+
+# Identify Power Button devices
+power_button_devices = get_power_button_devices()
```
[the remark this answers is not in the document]
knuton: Exactly

yfyf (approved):
Looks good, additionally tested manually that removing the `power-management` import in the test actually causes a failure.

PR #225, opened 2025-02-12
Enable mouse cursor hiding when idle

knuton:
This lets the mouse cursor be initially hidden, appear when moved, and hidden again after 10 seconds of being idle.

The timing is chosen so that the cursor is normally hidden, but stays visible for relatively long period should someone attach and use a mouse.

We have been hiding idle cursors in the Play web app, but have found that the mouse needs to be moved once upon boot for the web app's logic to work, which leaves the cursor permanently visible in mouse free use. This seems to be a system/engine level issue, that we can not control from the web app. By moving mouse hiding to the OS we can also remove the complexity and event handling overhead from the web runtime.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/application.nix
+++ b/application.nix
@@ -141,6 +141,13 @@ rec {
 
         defaultSession = sessionName;
       };
+      # Hide mouse cursor when not in use
+      services.unclutter-xfixes = {
+        enable = true;
+        timeout = 10;
+        threshold = 1;
+        extraOptions = [ "start-hidden" ];
+      };
 
       # Firewall configuration
       networking.firewall = {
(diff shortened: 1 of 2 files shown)


knuton:
> How does it work in addition with hiding the cursor in Play? I guess we should stop hiding the cursor in Play at some point?

Yes, please check #2288 in the apps repo, there's a link to a queued task for removing the functionality in the future. I wouldn't remove it from Play before most installations run the updated PlayOS.

PR #226, opened 2025-03-25
Set driver version to prohealth specific branch

yfyf:
Created a `studies/prohealth` based on `main`, used as a target here.

cc @stoeffel 

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated


knuton (approved):
Checked reference :+1:

PR #231, opened 2025-06-20
PlayOS network watchdog and proxy-utils

yfyf:
This PR introduces the network watchdog service and also extracts current proxy retrieval into a tiny `proxy-utils` package.

Extensive testing is done in the nixos test script, which I believe covers all the essential scenarios. The only missing case is maybe a multi-service setup, but I'm not sure it's relevant from the point of view the watchdog - it treats connman as a blackbox and does not really interpret the specific connection setup.

Also did basic manual tests with `./build vm` (with both the default watchdog config and the "faster" one added in 34c471e), confirmed that:
- watchdog is not doing anything when it can reach play.dividat.com
- watchdog tries to restart connman after the specified retries when I cut the internet connection
- watchdog detects a working internet connection after I restore it
- watchdog delays checks after changes in the controller UI
- watchdog uses proxy after its set in the controller UI

In theory it's possible [to also extract the connman DBus monitoring logic](https://github.com/dividat/playos/blob/main/kiosk/kiosk_browser/proxy.py#L125-L139) from [kiosk/kiosk_browser/proxy.py](https://github.com/dividat/playos/blob/main/kiosk/kiosk_browser/proxy.py) into `proxy-utils`, but I wanted to keep this changeset smaller, so decided against it.

## TODOs / open questions:

- [x] Perform more manual testing with multi-service setups?
- [x] Review logging levels / granularity, particularly in non-debug mode - should probably print how and which URLs failed?
- [X] Should we now disable the connman connectivity check? Or at least make it consistent by setting the primary URL of the watchdog?
    - Agreed not to disable, but review current settings and URL used for checking as a separate task 
- [x] Do we need to provide a way to disable the watchdog in the controller settings UI?
- [x] Squash commits before merge

## Manual testing scenarios

1. [x] Ethernet failover to wifi: connected to both, ethernet cable unplugged, connman switches to wifi, **watchdog does nothing**
2. [x] Wifi failover to ethernet: same as above, but reverse and cable unplugged is "AP is shut down".
3. [x] Wifi AP1 failover to AP2: same as above, but failover from one AP to the other
4. [x] Wifi AP temporary gone: connected to AP -> AP goes down -> watchdog restarts connman -> connectivity restored after AP reappears
5. ~~Connman connectivity check prioritizes ethernet over wifi: both connected -> initially wifi is prioritized over ethernet -> ethernet reaches state=online using connectivity check -> ethernet becomes prioritized over wifi. Expectation: watchdog does not break the prioritization.~~

## Checklist

-   [x] Changelog updated
-   [X] Code documented
-   [x] User manual updated


testing/integration/kiosk-screen-resolution.nix
```diff
@@ -82,8 +82,6 @@ pkgs.nixosTest {
 
   testScript = ''
     ${builtins.readFile ../helpers/nixos-test-script-helpers.py}
-    import time
```
yfyf, line 85: Side-note: it's probably time to upgrade `nixos-test-script-helpers.py` into a proper python package/module, to avoid having to do fixes like these.

base/networking/watchdog/watchdog.py
```diff
@@ -0,0 +1,265 @@
...
+    monitor.start_monitoring()
+
+    while True:
+        proxy = monitor.get_current_proxy()
+        url_check = make_url_checker(cfg, proxy)
+
+        state = override_state_if_connman_properties_changed(cfg, state, monitor.last_update)
+
+        debug(f"Current state: {state}")
+        match state:
```
yfyf, line 240: Was thinking to refactor this match/case dispatch into actions into a one-liner:
`state = state.run(cfg, url_check)`

by making the `run_state_X` a `run` method in the `StateX` classes (maybe via a base class), but then I thought it's kinda ugly mashing the data types with the actions and decided against it, even though this does add a bit more boiler plate.

pkgs/default.nix
```diff
@@ -26,6 +26,10 @@ let
       });
 
       connman = (import ./connman) super;
+
+      playos-proxy-utils = import ../proxy-utils {
```
yfyf, line 30: TODO: move to `pkgs.python3Packages` instead probably?

base/networking/watchdog/watchdog.py
```diff
@@ -0,0 +1,265 @@
...
+    return lambda: perform_url_checks(cfg.check_urls, cfg.check_url_timeout, proxy)
+
+
+def override_state_if_connman_properties_changed(cfg, current_state, last_update) -> State:
+    time_since_update = datetime.datetime.now() - last_update
+
+    if time_since_update.total_seconds() < cfg.setting_change_delay:
+        # override state, because there are recent connman changes
+        debug("Connman service properties changed recently, overriding current state.")
+        return StateSettingChangeDelay(time_since_update, next_state=current_state)
```
knuton, line 222: I find this more complicated to reason about than just storing the timestamp of the update as a datum, and computing the remaining time when processing the state.

Maybe there's some tradeoff I don't see?

knuton (changes requested):
Looks good!

I am going to do a second, deeper pass on the checking loop. I have some initial remarks about structure and details of check implementation.

Also installing on test machine now.

knuton:
I attached journalctl logs of a test installation to the ticket.

base/networking/watchdog/watchdog.py
```diff
@@ -0,0 +1,265 @@
...
+            'http': proxy.to_url(),
+            'https': proxy.to_url()
+        }
+
+    try:
+        # stream=True avoids downloading the content, which we don't care about
+        r = requests.get(url, headers=CLIENT_HEADERS, timeout=timeout,
+                         stream=True, allow_redirects=True, proxies=proxies)
+        if r.status_code != 200:
+            failure = RuntimeError(f"Bad HTTP status code: {r.status_code}")
```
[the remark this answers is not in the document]
yfyf: We could expand this to 2xx, but I would say, yes, anything but 2xx should count as error, since:
1. We follow redirects, so 3xx is handled
2. We control the URLs, so we don't expect them to return something weird
3. A network misconfiguration could lead to a broken DNS setup (that would probably fail at SSL though, unless a HTTP checkURL is used) or a proxy configuration that does not actually work (e.g. if connman falls back to some previously configured service which is no longer valid).

Unless we plan to expand the check URLs to 3rd party controlled endpoints I'd say let's stick with 2xx and redirects.

base/networking/watchdog/watchdog.py
```diff
@@ -0,0 +1,265 @@
...
+    except Exception as e:
+        failure = RuntimeError(f"Failed to connect: {e}")
```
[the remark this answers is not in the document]
yfyf: Just a cheap way to get a named type and avoid being all stringy everywhere. In theory they can be grouped into an `ExceptionGroup` and etc for clearer reporting. I'll review this when reviewing logging.

base/networking/watchdog/watchdog.py
yfyf: Hah, my purist foundations are leaking here :grin: 

In order to do what you suggest, the state handler would have to call `datetime.now()`, which would not match the `datetime.now()` used to invoke the SETTING_CHANGE_DELAY state (i.e. used in computing `time_since_update`), which is not great? Of course we can ignore this and compute `now()` twice, not an issue in this specific case.

An alternative would be to pass `remaining_delay` (`= cfg.setting_change_delay - time_since_update.total_seconds()`) instead of `time_since_update`, that would make the logic clearer maybe, but also moves more of the logic outside of the state handler.

base/networking/watchdog/watchdog.py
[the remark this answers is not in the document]
yfyf: Now accepting all 2xx and 3xx codes via `requests.raise_for_status()`: c030bd8b97d4e9

base/networking/watchdog/watchdog.py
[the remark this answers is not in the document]
yfyf: Removed in favour of a dataclass: 290468df78

yfyf:
> I attached journalctl logs of a test installation to the ticket.

Implemented a signal ignore list (that currently only contains `Strength` updates) in 301979ddd

Added not-perfect-but-better-than-nothing tests for two wifi scenarios, see [testing/integration/network-watchdog-wifi.nix](https://github.com/dividat/playos/blob/9323e50a2932f3c3a587e2689f1f5e705f9a3e3a/testing/integration/network-watchdog-wifi.nix)

pkgs/default.nix
yfyf: Done in eba2251

base/networking/watchdog/watchdog.py
```diff
@@ -0,0 +1,265 @@
...
+def log(msg):
+    logger.info(msg)
+
+
+def debug(msg):
+    logger.debug(msg)
+
+
+# Returns None if successful or error message if failed
+def perform_single_url_check(url, timeout, proxy=None):
```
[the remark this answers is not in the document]
yfyf: Done and later revised in 290468d

base/default.nix
```diff
@@ -49,6 +50,11 @@ with lib;
 
     playos = { inherit version kioskUrl; };
 
+    playos.networking.watchdog = {
+        enable = true;
+        checkURLs = [ kioskUrl "https://api.dividat.com" ];
```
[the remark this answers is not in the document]
knuton: We have to, because we should align with the firewall whitelist for the installation.

base/networking/watchdog/watchdog.py
```diff
@@ -0,0 +1,265 @@
...
+        debug(f"URL check for {url} failed with {failure}")
+    else:
+        debug(f"URL check for {url} succeeded!")
+
+    return failure
+
+
+def perform_url_checks(urls, timeout, proxy: proxy_utils.ProxyConf | None):
+    url_queue = deque(urls)
+    while url_queue:
```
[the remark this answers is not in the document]
knuton: I am happy to go with this, I am not sure the "simpler" temporal model would be worth the more complicated execution model.

base/networking/watchdog/watchdog.py
knuton: I think `remaining_delay` would have been more intuitive for me, but I get it now and am OK either way.

base/networking/watchdog/watchdog.py
knuton: Yeah, I think it's fine as is.

knuton (commented):
Very comprehensive and clearly structured tests! 👌 

My code review is complete with only a minor comment, renewed test installation with known volatile WiFi is up and running.

testing/integration/network-watchdog.nix
```diff
@@ -0,0 +1,338 @@
...
+    playos.wait_for_unit('always-ok-http-service.service')
+    playos.wait_for_unit("network-online.target")
+
+with TestPrecondition("PlayOS can reach HTTPStubServer`s"):
+    # assert we can reach the stub servers...
+    playos.succeed("curl ${primaryCheckUrl}")
+    playos.succeed("curl ${secondaryCheckUrl}")
+    # ...but they return an HTTP status code >=400
+    playos.fail("curl --fail ${primaryCheckUrl}")
+    playos.fail("curl --fail ${secondaryCheckUrl}")
```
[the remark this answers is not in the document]
yfyf: Do you mean the `fail(..)`? I'd say yes, it validates that the test setup is correct and watchdog gets non-OK status codes.

base/networking/watchdog/watchdog.py
yfyf: 33d184bacdd7e15ee3

yfyf:
> > * [ ]   Do we need to provide a way to disable the watchdog in the controller settings UI?
> 
> Nice to have, unless effort is higher than we expect.

Implemented in e0622e1.

In the UI it looks like this, except that I renamed the button labels in b845768:
image

base/networking/watchdog/default.nix
```diff
@@ -0,0 +1,120 @@
...
+
+  config = lib.mkIf cfg.enable {
+    systemd.services."playos-network-watchdog" = {
+      description = "PlayOS network watchdog";
+
+      after = [ "network.target" "connman.service" ];
+      wantedBy = [ "multi-user.target" ];
+
+      # allow to persistently disable using controller UI
+      unitConfig.ConditionPathExists = "!/home/play/.config/watchdog/disabled";
```
knuton, line 96: Shall we name it really like the service? It looks verbose, but it's clear.

testing/integration/controller-watchdog-controls.nix
```diff
@@ -0,0 +1,83 @@
...
+
+    playos.start()
+
+    playos.wait_for_unit("playos-controller.service")
+    playos.wait_until_succeeds("curl --fail http://localhost:3333/")
+    playos.wait_for_unit("playos-network-watchdog.service")
+
+    playos.succeed("systemctl is-active playos-network-watchdog.service")
+
+    with subtest("Network watchdog disable works"):
```
knuton, line 66: Should we include a test that it lives past a reboot?

yfyf: It's a bit tricky to do with nixOS test (you need to create disk image at test runtime for persistence) and it seems to break in funky ways when you start rebooting the VM (hence why I made these tests separate from the `controller-system-buttons.nix` tests).

yfyf:
> What seems notable are the setting changes, do we expect those? No manual action was taken.

I think they are expected, yes. You seem to be observing up to 4-5 updates per day, which can be due to updates initiated by DHCP, DNS, kernel or others.

> Should we succinctly print the kind of change even when not in debug level?

Sure, why not, might be helpful for debugging, I'll add it.

testing/integration/controller-watchdog-controls.nix
knuton: OK gotcha

Would it make any sense to break the abstraction layer and check that there is a file on disk?

yfyf:
Performed the manual tests on a live system with these extra modifications:

```
diff --git a/live/default.nix b/live/default.nix
index 308f6f3..fdbab1a 100644
--- a/live/default.nix
+++ b/live/default.nix
@@ -44,6 +44,17 @@ let nixos = pkgs.importFromNixos ""; in
       # There is no persistent state for a live system
       system.stateVersion = lib.trivial.release;
 
+      playos.xserver.activeVirtualTerminals = mkForce [ 1 7 8 ];
+      users.users.root.initialHashedPassword = "";
+
+      playos.networking.watchdog = {
+          enable = true;
+          checkURLs = [ kioskUrl ];
+          maxNumFailures = 3;
+          checkInterval = 10;
+          settingChangeDelay = 15;
+      };
+
     };
   };
   system = "x86_64-linux";
```

All of them worked as expected, except for the last one

> 5. [ ] Connman connectivity check prioritizes ethernet over wifi: both connected -> initially wifi is prioritized over ethernet -> ethernet reaches state=online using connectivity check -> ethernet becomes prioritized over wifi. Expectation: watchdog does not break the prioritization. 

which I could not really perform, since connman's connectivity check logic is a bit nonsensical, at least in the current "one-off" configuration we are using.

The scenario I tried:
- Connect PlayOS to a wifi which has no internet access at all
- Also connect via ethernet with a proper internet connection
- Both services remain in `state = ready` forever
- Manually disconnect from the wifi
- Ethernet service still remains in `state = ready`
- Re-plug ethernet service
- Ethernet service's `state` now actually becomes `online`

I was very confused as to why connman never transitions the ethernet connection to `online` (and hence never promotes it), but then I carefully re-read the manual and it says:

> In one-shot mode, there is a single, one-shot "online" HTTP-based Internet reachability check **for the default service.**

Since wifi is the default (and it is connected), no check for ethernet service ever happens! Not sure why it also doesn't happen when it _does_ become the default, but that's a different topic.

I also tried to analytically break-down the possible scenarios, but I don't believe there's any plausible scenario in which the watchdog might negatively interact with connman's connectivity check. 

I also did one more scenario in which the watchdog actually did something useful to recover the connection:
1. Connect to wifi and eth
2. (Temporarily) Kill the wifi AP
3. Connman switches to eth
4. Make the eth connection to the internet broken without disconnecting the cable
5. Connman remains stubbornly connected to eth
6. Watchdog detects the broken connection and restarts connman
7. Connman reconnects to wifi AP and internet is restored

yfyf:
Squashed, rebased, updated changelog and user manual. Ready for merge.

knuton (approved):

PR #234, opened 2025-07-10
Simulate a working self-update setup in ./build vm

yfyf:
This is achieved by slightly generalizing and re-using fake-rauc-boot and providing the stub server.

This was prompted by getting tired of mocking RAUC/controller to just see how the status page UI looks in the VM.


knuton (approved):
Thank you! :ok_hand:

PR #235, opened 2025-07-21
Add an S3 cache secondary substituter and post-build-hook for uploading

yfyf:
This was painful and took much longer due to the slow and obscure feedback loop of Github Actions. See comments/referenced issues for details.

Anyway, everything seems to work:
- builds are signed and uploaded: https://github.com/dividat/playos/actions/runs/16420048350/job/46395839724#step:7:6 (you can verify locally with e.g. `nix store verify --store https://dividat-ci-nix-cache.s3.eu-central-1.amazonaws.com/ /nix/store/pvnw37av6397kv3xwykda7my9n3k58cp-string-hosts` if you add the public key to `trusted-public-keys`)
- cache is accessible on forks for reads: https://github.com/yfyf/playos/actions
- large closures are being skipped: https://github.com/dividat/playos/actions/runs/16420048350/job/46395839729#step:4:4698

Regarding the last point - it would be preferable to skip not whole closures, but individual derivations/packages, but `nix copy` operates on closures, so I don't think that's possible.

### DIFF
--- a/.github/actions/prep-build-env/action.yml
+++ b/.github/actions/prep-build-env/action.yml
@@ -1,12 +1,85 @@
 name: "Prepare nix build environment"
+
+inputs:
+  NIX_CACHE_ACCESS_KEY_ID:
+    description: AWS S3 bucket access key ID
+    required: false
+
+  NIX_CACHE_SECRET_ACCESS_KEY:
+    description: AWS S3 bucket secret access key
+    required: false
+
+  NIX_BINARY_CACHE_PRIVATE_KEY:
+    description: private key used for signing store paths for the nix cache
+    required: false
+
+  # these are not intended to be configurable, but there's simply no way
+  # to have constants/envs in a composite action
+  NIX_CACHE_S3_BUCKET:
+    default: dividat-ci-nix-cache
+
+  NIX_CACHE_S3_PROFILE:
+    default: nixcache
+
+  NIX_CACHE_S3_REGION:
+    default: eu-central-1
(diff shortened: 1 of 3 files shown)


yfyf:
> EDIT: Hmm, maybe we can skip not based on the total size of the closure, but on the size of its (individual) packages/dependencies. This way we can distinguish between a "naturally" big closure (e.g. system image that is just made of many packages) and a e.g. a test closure that depends on a (huge) disk image.

Implemented this in 7bd441a

.github/actions/prep-build-env/action.yml
```diff
@@ -1,12 +1,85 @@
...
+      # current user and root, since it's tricky to ensure different nix
+      # commands (nix-build, nix copy) read from the right locations, see:
+      # https://github.com/NixOS/nix/issues/2161
+      sudo chmod -R a+rx /root/.aws
+      sudo chmod a+rx /root/
+      cp -r /root/.aws ~/.aws
+
+
+  - name: Install post-run action
+    uses: gacts/run-and-post-run@v1
```
knuton, line 69: Kind of nervous about this action dependency:

- Low usage
- GH org with unclear background and specific focus on providing GH Actions utils
- Gen AI illustration as parent org readme
- Verification "in progress" according to GH issue no 33

knuton: Thanks

knuton (approved):
:checkered_flag:

PR #239, opened 2025-09-04
Rely on link-local sensitive sorting to avoid using Senso as Internet gateway

knuton:
This does two things:

- Patch ConnMan to use a heuristic to demote IPv4 link-local services (sorts down Senso-links)
- Remove other ConnMan config that was supposed to avoid mistaking Senso-links for Internet gateways 

It relies on a ConnMan patch that has not been judged/accepted upstream (made a [mailing list post about the general idea](https://lore.kernel.org/connman/CAKQ2e6X9BVihgY58OjzwbaqNn5HVHqHDCirsO2+qCBro91sygg@mail.gmail.com/T/#u)), though I am slowly gaining confidence in the approach being sensible. We could drop the patch and replace with another method with similar effect if there is no interest to integrate upstream.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -6,6 +6,10 @@
 - kiosk: Add virtual keyboard
 - kiosk: Auto-disable virtual keyboard when real keyboard detected
 
+## Fixed
+
+- os: Prevent the network manager from attempting to route Internet traffic to the Senso in networks with delayed DHCP
+
 # [2025.3.1] - 2025-07-04
 
 # [2025.3.1-VALIDATION] - 2025-06-06


pkgs/connman/ipconfig-method-in-sorting.patch
```diff
@@ -0,0 +1,140 @@
...
++ *  @sa service_list_sort
++ *  @sa start_online_check
++ *  @sa __connman_connection_update_gateway
++ *  @sa __connman_timeserver_sync
++ *
++ */
+ static void address_updated(struct connman_service *service,
+ 			enum connman_ipconfig_type type)
+ {
++	service_list_sort();
```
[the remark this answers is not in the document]
knuton: That was introduced after v1.42: https://git.kernel.org/pub/scm/network/connman/connman.git/tree/src/service.c?h=1.42

pkgs/connman/ipconfig-method-in-sorting.patch
```diff
@@ -0,0 +1,140 @@
...
++		return 1;
++
+ 	if (service_a->favorite && !service_b->favorite)
+ 		return -1;
+ 
+@@ -7124,6 +7177,7 @@ static void service_ip_release(struct connman_ipconfig *ipconfig,
+ 					CONNMAN_IPCONFIG_TYPE_IPV4);
+ 
+ 	settings_changed(service, ipconfig);
++	address_updated(service, type);
```
yfyf, line 137: Is this needed for the nixos test to pass? I would guess that not, but curious if my intuition is wrong. This seems like a sensible change, but slightly orthogonal. Were you thinking of the Aterlier 1 issue with random IP addr removals? We can actually create another simple test for such a scenario (service/interface gets IP removed and we would prefer connman to respond to that).

pkgs/connman/ipconfig-method-in-sorting.patch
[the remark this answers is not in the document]
yfyf: Oh, sorry, was looking at connman 1.44 sources, nevermind.

pkgs/connman/ipconfig-method-in-sorting.patch
knuton: No, not necessary for the test to pass and I added it only "for symmetry" let's say.

This is one of the details I am not sure about yet. Practically sorting would only be affected if the ipconfig _method_ has changed, not from address removal alone.

yfyf: We can check, but I think it will fail on both versions, because neither of them respond to IP address removals? With the Atelier1 scenario it was more of a question of what _triggered_ the IP removal in 1.43.

knuton: Mhh yes, that was the mystery.

yfyf:
I think the patch is probably fine, but... I was trying to understand the broader impact of this change and how could we test that this doesn't introduce new types of bugs. 

A quick survey on connman 1.42 (without this patch):

Both `service_list_sort` and `__connman_connection_update_gateway` are called when:
- favorite services change (`__connman_service_set_favorite_delayed` and `__connman_service_provision_changed`)
- service reaches an error state (`report_error_cb`)
- service changes state (`service_indicate_state`)
- new service is registered (`service_register`)

Only `__connman_connection_update_gateway` is called on:
- a service move (`__connman_service_move`)
- service removal (`service_free`)
- user-input during configuration (`request_input_cb`)
- preferred service order change (`service_update_preferred_order`)
- service is created "from network" ( `__connman_service_create_from_network`)

 Only `service_list_sort` is called on:
- update from network (`update_from_network` and `__connman_service_update_from_network`)

To me this reeks of potential bugs where the default gateway is not updated in sync with the top-sorted "default" service. It is really weird that e.g. `__connman_service_move` is not re-used in the sorting implementation itself or that there are gateway updates during the handling of user input (e.g. why does `report_error_cb` resort and update GW while `request_input_cb` only updates GW?).

pkgs/connman/ipconfig-method-in-sorting.patch
yfyf: > Practically sorting would only be affected if the ipconfig _method_ has changed, not from address removal alone.

Speaking of which, is that the reason you look at the method and not the actual address in `is_ipv4ll`? `acd.c` already defines a predicate `is_link_local` that could be (re-)used to check that the IP is link-local. I'm not sure it makes sense to use the sorting criterion based on method, when... there's no address (i.e. after removal)?

knuton:
Thanks for assembling the cases. In general I think we look at the landscape similarly. Some details I am not sure I agree on/understand.

> It is really weird that e.g. `__connman_service_move` is not re-used in the sorting implementation itself

Why? This is for imperative moving of services, the list sorting is just using a library function and the custom comparison function.

> Both `service_list_sort` and `__connman_connection_update_gateway` are called when:

These seem to be cases where service properties change in a way that can change ordering relative to several other services.

> Only `__connman_connection_update_gateway` is called on:

These should be cases where a service is just removed or swapped directly on user request. The resulting order is then clear without sorting again.

I likewise don't understand why `request_input_cb` and `__connman_service_create_from_network` are in this category.

> Only `service_list_sort` is called on:
> 
> * update from network (`update_from_network` and `__connman_service_update_from_network`)

This does not make sense to me either. One might think that omitting the gateway change here is a bit of a low pass filter, because e.g. wifi signal strength changes often. But then why even sort and not wait for another sorting trigger as well?

pkgs/connman/ipconfig-method-in-sorting.patch
knuton: Here's a slight rewrite of the sorting block to make the assumptions clearer (behavior should be identical).

```diff
diff --git i/src/service.c w/src/service.c
index 44884fff..7bf8b10b 100644
--- i/src/service.c
+++ w/src/service.c
@@ -5463,13 +5463,21 @@ static gint service_compare(gconstpointer a, gconstpointer b)
 			return 1;
 	}
 
-	/* Prefer configured IPv4 address over auto/link-local IP. */
-	bool a_is_link_local = is_ipv4ll(service_a);
-	bool b_is_link_local = is_ipv4ll(service_b);
-	if (a_connected && !a_is_link_local && b_is_link_local)
-		return -1;
-	if (a_is_link_local && b_connected && !b_is_link_local)
-		return 1;
+	/* Prefer configured IPv4 address over auto/link-local IP.
+	 *
+	 * Online services would already have won against ready.
+	 *
+	 */
+	if (state_a == CONNMAN_SERVICE_STATE_READY &&
+			state_b == CONNMAN_SERVICE_STATE_READY) {
+		bool a_is_link_local = is_ipv4ll(service_a);
+		bool b_is_link_local = is_ipv4ll(service_b);
+
+		if (!a_is_link_local && b_is_link_local)
+			return -1;
+		if (a_is_link_local && !b_is_link_local)
+			return 1;
+	}
 
 	if (service_a->favorite && !service_b->favorite)
 		return -1;
```

If one of the services is _online_, it would already have won. If one of the services is _ready_ and the other is not, the _ready_ service would already have won.

So we only apply the rule if both services are _ready_ (the rewritten version makes this explicit and less dependent on ordering of the comparison blocks), and then let a non-link-local service win over a link-local service.

We could look at IP addresses, but a service should not be _ready_ without an IP address, which is why I thought method would be enough.

base/networking/default.nix
```diff
@@ -33,14 +33,18 @@ in
...
-      PreferredTechnologies=wifi,ethernet
-
       # Allow simultaneous connection to ethernet and wifi
       SingleConnectedTechnology=false
 
-      # Enable online check to favour connected services
-      EnableOnlineCheck=true
+      # Always use WiFi and Ethernet in parallel
+      AlwaysConnectedTechnologies=ethernet,wifi
+      DefaultAutoConnectTechnologies=ethernet,wifi
```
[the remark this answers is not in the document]
knuton: Done

base/networking/default.nix
```diff
@@ -33,14 +33,18 @@ in
...
+      # Use Ethernet without active user prompt (plugging cable is the prompt)
+      DefaultFavoriteTechnologies=ethernet
```
yfyf, line 43: This also matches the default value, do we need to explicitly set it? The diff makes it seem like we are "changing" something, whereas in reality this has always been configured this way.

yfyf (approved):
I think we are ready to go with this. Can merge as is, but I would remove the configuration options which remain the same as defaults.

base/networking/default.nix
yfyf: That's fine, I was just confused about the diff, thinking that these settings are "new"

PR #240, opened 2025-09-04
Fix loop in controller dev-server

knuton:
watchexec started reacting to watch command outputs for me, executing the watch command in a rapid loop. Possibly this is due to a change/bug in how the gitignore is processed, as keeping a gitignore in the project subfolder with local paths fixes the issue.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,3 @@
 result
 work/
 release/
-*.merlin
-*.install
-*/_build
-controller/Changelog.html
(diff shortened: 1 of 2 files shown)


yfyf (approved):
Looks good, I actually prefer per-component .gitignores instead of one top-level .gitignore, so this makes sense.

PR #241, opened 2025-09-04
Make network interface getter more robust

knuton:
On my dev machine, the network overview page could not be opened (just displayed a `Not found`). The reason turned out to be that there is an LTE modem in my laptop (interface name `wwan0`), for which `ip -j link` did not report an `address` field, and so parsing failed with an exception.

As we are practically only interested in interface types which always have a link layer address, and other types can not even be configured through our UI, skip entries which do not parse into our expected format.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/server/network.ml
+++ b/controller/server/network.ml
@@ -68,13 +68,23 @@ module Interface = struct
     )
 
   let of_json j =
-    let dict = Ezjsonm.get_dict j in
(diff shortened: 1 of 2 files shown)


yfyf (approved):
Looks totally reasonable, thanks!

PR #247, opened 2025-09-16
Remove our S3 substituter for now

knuton:
The egress cost is higher than anticipated, disable until cost/options could be analyzed.

Added in https://github.com/dividat/playos/pull/235

If we disable this and egress remains at a similar level, we must assume there is another party causing the traffic.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


yfyf:
> another party causing the traffic

I was one other party, I had it set up (read-only) on my laptop too. Don't think I could have managed to do much traffic though.

PR #251, opened 2025-10-03
Replace S3 with R2 bucket

yfyf:
Making the substituter use private credentials for now, since we need to configure a custom domain for R2 for public access. Will do this later.

Kind of defensively specifying everything multiple times, because it seems that nix fails to read some parameters (e.g. the `endpoint_url`) from the aws profile configuration. 

Did some test runs, seems to work:
- https://github.com/dividat/playos/actions/runs/18221865628
- https://github.com/dividat/playos/actions/runs/18222050137 (flaky watchdog test, need to look into it)

Setup Cloudflare usage notifications for the R2 bucket with quite conservative limits:
- Storage exceeds 200GB (== 2$/month)
- Class A or B operations exceed 3M per month (== 15$/month)

Having an actual hard billing cap is impossible, it seems :(

Also set up a lifecycle rule to delete objects 1 year (365 days) after creation.


knuton (approved):
Config looks good. Let's see it in action!

PR #263, opened 2025-10-27
Disable power_save feature on any WiFi cards

knuton:
Adds a service to disable power saving for all WiFi devices on boot.

Unfortunately power saving has been implicated in random disconnects for some of our hardware configurations and customer APs. In simple tests observing consumed energy of an idle system using a Watt meter, there was no meaningful difference when power saving was disabled.

We disable power saving for all detected WiFi devices without offering a documented way for the application layer to override, assuming any installations are stationary and mains powered. Powersaving behavior could be made configurable if use cases evolve going forward.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -1,3 +1,7 @@
+## Changed
+
+- os: Disable WiFi power-saving mode for increased network robustness
+
 ## Fixed
 
 - os: Prevent the network manager from attempting to route Internet traffic to the Senso in networks with delayed DHCP


knuton:
This seems like a good measure for our mains-powered installations, even though it is just one of many possible parameters affecting connection robustness.

I am a bit on the fence whether to include in a patch release.

@yfyf Would appreciate your input and review

PR #264, opened 2025-10-29
 Enable spatial navigation into and within Qt elements with arrow keys

yfyf:
...and also deal with a bunch orthogonal annoyances, including:

- Make kiosk button styling uniform and focusing more visible
- Prevent kiosk buttons from handling spatial navigation (needed to avoid wrapping)
- Remove the multi-action dialog that is redundant now

As usual with Qt, this turned out to be quite elaborate and quite silly, but is the best I can come up with. It should be fairly robust to various re-arrangements and already handles situations like "double stacking" when in PlayOS settings (i.e. dialog) and captive portal message is visible.

I tried and/or considered several different approaches, but they seemed even messier:
-  "manual" focus in/out handling in BrowserWidget - this turned out to be a mixture of the current approach and extra manual handling, so is even harder to grok, leaks the logic across multiple widgets
-  "marking" widgets as "top" / "bottom" explicitly and having them override `focusNextPrevChild` behaviour to prevent wrapping - this seemed way more cumbersome and requires to track/propagate info about the "top" widget
- translating arrow key presses into Tab / Shift+Tab - this avoids all the boiler plate code, but it means navigation is very unpredictable spatially

Remaining questions/issues, a bit orthogonal:
- If wrapping is acceptable, then this can be greatly simplified, but I think wrapping sucks?
- The button styling is a bit crude, the border is not very visible on a light background.
- The `focus-shift:exhausted` events do not trigger on many pages I tried, making it impossible to focus on the close button in Network login (captive portals).
- There are some issues with double-execution of focus-shift in some strange pages with sub-frames, we can look at this separately.

## Checklist
-   [ ] ~~Changelog updated~~
-   [X] Code documented
-   [ ] ~~User manual updated~~

### DIFF
--- a/kiosk/kiosk_browser/browser_widget.py
+++ b/kiosk/kiosk_browser/browser_widget.py
@@ -108,8 +108,6 @@ class BrowserWidget(QtWidgets.QWidget):
         # Prevent opening context menu on right click or pressing menu
         self._webview.setContextMenuPolicy(QtCore.Qt.ContextMenuPolicy.NoContextMenu)
 
-        self.setFocusProxy(self._webview)
-
         # Prepare reload timer
         self._reload_timer = QtCore.QTimer(self)
         self._reload_timer.setSingleShot(True)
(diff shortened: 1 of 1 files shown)


knuton:
Looks good, yet to perform practical tests.

yfyf:
> the captive portal nagbar comes up immediately but I can not initially move focus to it with the arrow keys (tabbing works). 

Good catch.

> This might just be because the loading page does not have focus-shift?

Yes, most likely. I will do some testing with an empty page and an infinitely loading page to check what can be done.

> When I focus another window and then activate the kiosk again, the captive portal button is focused.

Classic scenario :skull:

PR #269, opened 2025-11-12
Apply patches to apply WiFi retry mechanism in more cases

knuton:
This extends the existing retry mechanisms to apply in more cases. Beyond this, it may be wise to bump up the number of allowed retries, which used to be 2 for unknown and 4 for known/favourite wifis originally (altered in
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=289c7fc5c).

While I think this patch has a chance to be accepted upstream, the bump of maximum retries may not.

This is the new patch to ConnMan (not already on `main`) that I hope will improve stability of OK-but-not-great-signal deployments of PlayOS.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -1,6 +1,8 @@
 ## Changed
 
 - os: Optimize the WiFi scanning strategy for stationary clients
+- os: Enable retries when connecting to WPA3 WiFi networks fails
+- os: Increase the number of permitted retries for failed WiFi connections
 
 ## Fixed
 


yfyf:
Do you intend to only merge this to main or also the release branch?

Any chance of creating a nixOS test, at least for checking if SAE auth gets retried on invalid-key? I have only briefly scanned the implementation, but my understanding is that this would retry even if the key is "actually" invalid?

I would be in favour of bumping the retries, even significantly to like 10, seeing how other vendors have patches for "unlimited" retries.

pkgs/connman/generic-wifi-retry.patch
```diff
@@ -0,0 +1,228 @@
...
+From: Johannes Emerich <johannes@dividat.com>
+Date: Wed, 12 Nov 2025 11:25:55 +0100
+Subject: [PATCH 1/3] Extend auth retry mechanism to WPA3-SAE
+
+Special treatment of WPA3-SAE was added in
+0d5d05f2a6a6895aa2b1ffeda34489552468327b with the explicit purpose of
+triggering a renewed password prompt by setting an invalid-key error if
+a disconnect occurred during SAE authentication. This was modeled on
+handling of 4-way handshake failures, replicating some but not all of
+the handler code.
```
yfyf: Maybe mention which "handler code" you refer to? You mean `handle_4way_handshake_failure`, right?

pkgs/connman/generic-wifi-retry.patch
```diff
@@ -0,0 +1,228 @@
...
++ * - Authentication transaction sequence number out of expected sequence
++ *
++ */
+ static bool has_transient_assoc_failure(struct wifi_data *wifi)
+ {
+ 	return ((wifi->state == G_SUPPLICANT_STATE_ASSOCIATING) &&
+-			wifi->assoc_code == ASSOC_STATUS_AUTH_TIMEOUT);
++			(!wifi->assoc_code ||
++				wifi->assoc_code == ASSOC_STATUS_OUT_OF_SEQUENCE ||
++				wifi->assoc_code == ASSOC_STATUS_AUTH_TIMEOUT));
```
yfyf: This is beyond my knowledge of 802.11, but can you explain how you arrived at this extra condition? Might be a good idea to include it in the commit message too.

pkgs/connman/generic-wifi-retry.patch
```diff
@@ -0,0 +1,228 @@
...
+0d5d05f2a6a6895aa2b1ffeda34489552468327b with the explicit purpose of
+triggering a renewed password prompt by setting an invalid-key error if
+a disconnect occurred during SAE authentication. This was modeled on
+handling of 4-way handshake failures, replicating some but not all of
+the handler code.
+
+We have observed frequent disconnects from WPA3 networks with ConnMan
+and wpa_supplicant as well as iwd, resulting in wifi services being
+marked with invalid-key error and not considered for connection until
+the system is rebooted or service manually selected again. These may be
```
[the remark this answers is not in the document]
knuton: Rephrased in 5512a16

knuton:
I also included bumps to maximum retries in the patch now, 2 for everything, 5 for previously successful configured services (favorites).

yfyf:
> Yes, I would like to extend/add VM tests for this.

Within the scope of this PR or afterwards?

The PR is ready for merging otherwise, but I think it would be preferable to have tests verifying that:
- SAE auth is retried (even in a silly way - just asserts on supplicant logs)
- deauth during assoc is retried (more tricky, but maybe we can think of something)

yfyf:
testing/release-validation build failed because disk ran out of space. It seems (some ?) Github runners have smaller disks or our builds we have grown in size. Made [an issue to look at it](https://github.com/dividat/playos/issues/271).

testing/integration/connman-wifi-retries.nix
```diff
@@ -0,0 +1,219 @@
...
+  updateUrl = "http://localhost:9000/";
+  # url where kiosk points
+  kioskUrl = "https://play.dividat.com";
+
+  fullProductName = "playos";
+  safeProductName = "playos";
+  version = "1.1.1-TEST";
+  greeting = s: "hello: ${s}";
+
+  playos-controller = import ../../controller {
```
yfyf: Do you really want to use controller here? If we are just testing the patch, then plain `connman` is easier to test and debug I guess?

testing/integration/connman-wifi-retries.nix
```diff
@@ -0,0 +1,219 @@
...
+        playos,
+        regex="can still retry wifi",
+        unit="connman.service",
+        timeout=45
+    )
+    # This log line is our proxy for having retried the connection a second time
+    wait_for_logs(
+        playos,
+        regex="BSSID 02:00:00:00:00:00 ignore list count incremented to 2",
+        timeout=45
```
yfyf: Should this specify a `unit=`?

PR #270, opened 2025-11-14
Apply a ConnMan patch for less aggressive bgscan

knuton:
This overrides the hardcoded bgscan setting with less aggressive values:

- Short scan period: Remains at 30 seconds
- Signal strength threshold: -70 dB, sufficient for basic web usage
- Long scan period: Much higher at 1 hour

The signal strength threshold is chosen low at -70 dB, which should
still be tolerable but less than ideal for our kiosk applications. It
seems like a good choice based on this reasoning:

- A stable connection is higher priority than ideal throughput
- Both client and APs are assumed to be stationary
- The client will choose a suitable AP when first connecting

See the patch for more details.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -1,3 +1,7 @@
+## Changed
+
+- os: Optimize the WiFi scanning strategy for stationary clients
+
 ## Fixed
 
 - os: Prevent the network manager from attempting to route Internet traffic to the Senso in networks with delayed DHCP
--- a/pkgs/connman/default.nix
+++ b/pkgs/connman/default.nix
@@ -21,6 +21,10 @@ super.connman.overrideAttrs (old: {
         url = "https://git.kernel.org/pub/scm/network/connman/connman.git/patch/?id=d90b911f6760959bdf1393c39fe8d1118315490f";
         hash = "sha256-odkjYC/iM6dTIJx2WM/KKotXdTtgv8NMFNJMzx5+YU4=";
       })
+      # Custom patch to use more appropriate bgscan settings for
+      # stationary clients. Will never be accepted upstream, but
+      # is minimal and easily updated.
+      ./stationary-bgscan.patch
       # Custom patch to add a ipconfig method heuristic to ConnMan,
       # can be made to work up to ConnMan 1.45 at least. If not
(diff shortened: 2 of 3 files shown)


knuton:
> Can you provide some background for this? The -65 to -70dB is fine, but the long bgscan of 1 hour is a bit... wild?

It doesn't seem that wild to me. [NetworkManager's default is 24 hours](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/1.54.1/src/core/supplicant/nm-supplicant-config.c?ref_type=tags#L735).

It does use the same parameters as ConnMan however when it sees a multi-AP Extended Service Set. I tried to find a full account of why, but it's all a bit sparse:

- https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/0f3203338c858115149581e090d443f4c4d33267
- https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/279
- https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/248

Here's some ancient more detailed background discussion: https://bugzilla.gnome.org/show_bug.cgi?id=766482

I was looking for a statement like "In a multi-AP environment, we assume that clients may be laptops in an office and might roam." or "In a multi-AP environment, we assume that APs may come in and out over time.", but didn't find anything quite so explicit.

> It can introduce issues in scenarios where a better AP temporarily becomes noisy or is shut down. PlayOS will jump to the bad AP for a whole hour, and at -70dB it can be quite bad indeed.

Wouldn't that be an argument that -70 dB is too low, rather than the period being too long?

> What is the purpose of reducing the frequency that much? My understanding is that this only helps with power saving, which is not a very relevant concern for us.

Two-fold:

- According to my understanding, each scan disrupts regular wifi communication, which adds to latency and I suspect may sometimes cause misunderstandings between client and AP. I don't find authoritative information for the specifics, 
- I want the system to stay connected with a decent AP rather than trying to optimize connection by switching APs. If each connection attempt with an AP has a 2 % chance of failing, then it matters whether the system does 3 or 34 AP switches a day.

Issue #271, opened 2025-11-17
testing/release-validation.nix needs more disk space on CI?


yfyf:
[Job logs](https://productionresultssa12.blob.core.windows.net/actions-results/4f6cd0f4-c8b5-401c-8afc-9173801ceab4/workflow-job-run-7207cf4c-91a8-5ddf-86b0-1f795d766d64/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-11-17T07%3A26%3A22Z&sig=yo50gT1p5wjnUzeKKw6fY%2BKNB26k2sgg%2BsxsKcsjK%2FA%3D&ske=2025-11-17T16%3A11%3A16Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-11-17T04%3A11%3A16Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-11-17T07%3A16%3A17Z&sv=2025-11-05) 

```
2025-11-17T06:30:21.5015640Z building '/nix/store/17y621124rnpwqb8rrnm2zslazwqml6x-user-units.drv'...
2025-11-17T06:30:21.6648050Z building '/nix/store/dskmx579kqdm3ycmmw8r6w7hkps1hij7-system-units.drv'...
2025-11-17T06:30:22.6724665Z building '/nix/store/193j0pqj8q342lmld07l1pvkci522i7a-etc.drv'...
2025-11-17T06:30:23.3190697Z building '/nix/store/p5df66h8pdf44g98ss566kfz46dby22a-nixos-system-playos-playos-9999.99.99.drv'...
2025-11-17T06:30:24.9357260Z building '/nix/store/96mgv6hb354px518rzg0gqzw0h10zpcv-closure-info.drv'...
2025-11-17T06:30:24.9462070Z structuredAttrs is enabled
2025-11-17T06:30:25.2245962Z building '/nix/store/rs64b89q0ixwrl7cy0kjzz3fypwiqxpg-bundle-9999.99.99.raucb.drv'...
2025-11-17T06:36:02.2156269Z ##[warning]You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 33 MB
2025-11-17T06:36:03.7457938Z Error writing block data
2025-11-17T06:36:03.7795295Z tar: -: Wrote only 4096 of 10240 bytes
[...]

PR #272, opened 2025-11-19
Add step to remove unused libraries to gain space

knuton:
We started seeing failures in the release-validation build on GitHub Actions due to full disk, but in PRs that should not materially impact size of artifacts at all. (https://github.com/dividat/playos/issues/271)

As a sanity check, started unmodified https://github.com/dividat/playos/actions/runs/19498481785 which I expect to also fail in the same way (confirming there is no change in the outputs we produce, but a change in Ubuntu runners).

This PR tries to free up space on the runner before starting the actual build.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


yfyf:
Nice, thanks!

PR #282, opened 2025-12-01
Set Intel wifi cards to always-active

knuton:
This adds kernel module parameters for the Intel wifi drivers to disable
power saving and set the power management scheme to "active"
permanently.

By default the submodule for most modern Intel cards, iwlmvm runs with a
"balanced" power management setting. This is suitable for
battery-powered devices, but often implicated in connection drops
(though depending also on the AP). In our stationary scenario,
connection stability is higher priority than reduced network card energy
consumption.

Power saving is disabled at two levels:

1. With a udev rule that should disable power saving for any WiFi card
   whose drivers support it.
2. Explicitly for Intel wifi drivers, setting the desired power scheme
   directly, should disabling power-saving not automatically do so.

We assume that the driver level would be sufficient for a regular
installation, but the udev rule seems like a low effort failsafe for
unusual setups.

We don't assume later changes to interfaces due to higher level
components like NetworkManager, which is not used.

An earlier and abandoned approach used a one-shot systemd service before
network.target, operating on all wireless interfaces
(https://github.com/dividat/playos/pull/263).

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -11,6 +11,7 @@
 ## Changed
 
 - controller: Enable spatial navigation to toggle visibility in password inputs
+- os: Disable power-saving on WiFi cards for greater connection stability
 
 ## Fixed
 


knuton:
Made it a bit less simple, but at least as effective. :-)

I figured the udev rule is simple enough, so I'd rather also disable power-saving for any non-Intel cards, should they slip in.

Tested the udev rule by setting the module params to enable power-saving/balanced scheme, and confirmed that it still resulted in power management disabled.

Also moved it to the base layer, because at least as long as PlayOS is assumed to be stationary/mains powered, this is not an application-level concern.

PR #283, opened 2025-12-04
Add release validation image for 2025.3.2

knuton:
PR while validation on added image still runs.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated


knuton:
@yfyf Is there a recommended way to set the sha256 hash for the image?

PR #287, opened 2025-12-08
Drop unused dependencies

knuton:
Try to drop unused dependencies for smaller bundle.

- 1d6bead17b37bee61cd6ee1e5a1a7f791739d060 seems unproblematic and desirable
- abb9617f8e73f65fe5165d3c388a769598df42e3 is probably also a good idea, but probably too dramatic for a patch release

I also tried

```diff
diff --git a/kiosk/default.nix b/kiosk/default.nix
index 60a8f3fb..2ecd8b6a 100644
--- a/kiosk/default.nix
+++ b/kiosk/default.nix
@@ -31,7 +31,7 @@ python3Packages.buildPythonApplication rec {
     pygobject3
     pyqt6-webengine
     pytest
-    qt6.qtbase
+    (qt6.qtbase.override { cups = null; })
     requests
     types-requests
   ] ++ additional_inputs;
```
but it doesn't shave of as much as I thought, while requiring Qt recompilation.

## Analysis techniques

- https://nixcademy.com/posts/minimizing-nixos-images/
  - `nix-store --query --requisites $(nix-build -A components.systemImage)   | xargs du -sm | sort -n`
  - `nix why-depends /nix/store/path-a /nix/store/path-b`
- `nix-tree $(nix-build -A components.systemImage)`

## Checklist

-   [ ] Changelog updated
-   [ ] Code documented
-   [ ] User manual updated

### DIFF
--- a/application/trim.nix
+++ b/application/trim.nix
@@ -29,9 +29,4 @@
       extraPortals = lib.mkForce [];
     };
   };
-
-  # Only include a single fallback font, assume kiosk app brings fonts
-  fonts.enableDefaultPackages = false;
-  fonts.fontconfig.enable = lib.mkForce false;
-  fonts.packages = lib.mkForce [ pkgs.fira ];
 }


yfyf:
Did you consider also adopting all or parts of [nixpkgs/nixos/modules/profiles/minimal.nix](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/profiles/minimal.nix)?

yfyf:
How much does this shave off? I kinda like `nix.enable = false`, given that PlayOS is meant to be 100% immutable.

knuton:
With the latest push it's about 70 MB, so not terribly amazing, but not bad either. Several of these make sense independent of specific space savings, too.

Need to think about whether it's worth including in the patch release, though.

```
# trimmed
> nix-build -A components.unsignedRaucBundle && ls -lL result
-r--r--r-- 1 root root 1307486436 Jan  1  1970 result

# release branch
> nix-build -A components.unsignedRaucBundle && ls -lL result
-r--r--r-- 1 root root 1379698916 Jan  1  1970 result
```

yfyf:
Changes look good and nicely set up, but I would say a 70MB difference is not worth even the tiniest risk that this might break something. I vote for merging this into `main` instead and releasing later as a regular release. Need to minimize moving parts for now.

PR #288, opened 2025-12-09
Download RAUC bundles to persistent storage

yfyf:
See the [task on Notion](https://www.notion.so/dividat/Download-RAUC-bundles-into-non-tmpfs-2c36ed7e605280919705d628616532f1) for an analysis of current persistent storage usage.

Although we have >2GB free space on the data partition even with the conservative estimates, I think it's saner to ensure there's more wiggle room for the future and to be more explicit about it. With these restrictions, we are guaranteed to have > 3.5GB of free space, which allows for bundle growth and unexpected additional usage.

The Driver patches to reduce logging can be handled in a regular (non-patch) release, which should "restore" the journald reductions in this PR.

Tested manually that Qt/Chromium respects the cache limits.

## Checklist

-   [x] Changelog updated
-   [X] Code documented
-   [X] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -4,6 +4,7 @@
 
 - controller: Extend wait after failed update installation attempt from 30 s to 5 min
 - controller: Delete system update files immediately after successful installation
+- controller: Download system update bundles to disk instead of RAM
 
 ## Fixed
 


base/default.nix
```diff
@@ -62,11 +62,25 @@ with lib;
         User = "root";
         RestartSec = "10s";
         Restart = "always";
+        ExecStartPre = "-${pkgs.writeShellScript "clear-state" ''
+            rm -rf /var/lib/playos-controller/*
+            rm -rf /var/lib/playos-controller/.*
+        ''}";
+        StateDirectory = "playos-controller";
```
knuton: Is this the contractual way?

I would have expected use of [`systemd.service.playos-controller.environment`](https://search.nixos.org/options?channel=25.05&show=systemd.services.%3Cname%3E.environment).

kiosk/bin/kiosk-browser
```diff
@@ -6,6 +6,7 @@ import os
 import kiosk_browser
 
 default_toggle_settings_key = 'CTRL+SHIFT+F12'
+default_max_cache_size = 1024*1024*100 # 100MB
```
knuton: Shall we put this a bit higher? This may in principle mean game assets need to be evicted from cache during dev use when cache limit is not explicitly set (though unlikely that someone loads the majority of assets), which might alter perceived site performance.

knuton (commented):
Thanks, this is great!

A few details to discuss.

base/default.nix
yfyf: I think so. This specifies that the service stores persistent data in `/var/lib/playos-controller`. This ensures the directory exists, is owned by the service user, etc. In the current setup, most of these are already "given", but this allows to change it seamlessly and communicates the intent.

base/default.nix
```diff
@@ -62,11 +62,25 @@ with lib;
         User = "root";
         RestartSec = "10s";
         Restart = "always";
+        ExecStartPre = "-${pkgs.writeShellScript "clear-state" ''
+            rm -rf /var/lib/playos-controller/*
+            rm -rf /var/lib/playos-controller/.*
+        ''}";
```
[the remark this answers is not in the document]
yfyf: I don't think so. The not covered case is probably meh, since it will be eventually cleaned upon reboot anyway.

I do wonder if we should add a back-off to the restart delay (via [RestartSteps=](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#RestartSteps=) and `RestartMaxDelaySec=`), because now it's (very unlikely, but possible) that e.g.

- controller downloads a new bundle
- something causes a controller crash, e.g. mid-download or during install
- service gets restarted with the current 10s delay
- state folder is cleared pre-start
- controller starts downloading the bundle from scratch
- repeat every 10 seconds

This can cause various issues at scale (network overload, costs, etc).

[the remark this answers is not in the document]
knuton: Mhhh, yes, but leaving the system without a controller is also no good.

To the extent that we worry about/observe this, we could [build defenses into the dist host](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/WAF-one-click-rate-limiting.html) to deny bursts of requests. In this way we could tell the client to back off, which would both reduce egress strain and possibly keep the client from overloading itself -- but without putting the burden on the client to live without controller.

By the way, I had the thought that it would be excellent if the update client would identify itself and PlayOS version to the dist server via the user-agent header.

kiosk/bin/kiosk-browser
knuton: OK, let's set this low, while possibly ramping up production cap.

application.nix
```diff
@@ -21,6 +21,9 @@ rec {
       (import ./application/overlays version)
     ];
 
+
+    max-browser-cache-size = 1024*1024*200; # 200MB, in bytes, not including profile
```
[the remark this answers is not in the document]
yfyf: Bump to 250 and logged possible historical value in the commit message: 8bd4a5f9a

PR #293, opened 2025-12-12
PoC: Move RAUC statusfile to persistent storage and check Status in release-validation post update

yfyf:
This started out with b9143cc - after the select-display bug, I wanted to extend the tests to check that post-update the system reaches a Good state. When I tried that, I realized that does NOT happen in release-validation tests, since `/boot/status.ini` gets corrupted and statusfile-recovery fails. This is due FAT shenanigans to the unclean `system_reset` used in the test. See attached screenshots.

So this is actually 2 PRs, but I wanted to illustrate that the moving of RAUC to persistent storage kinda works already in the tests.

The actual implementation can be simplified and needs more tests/clean-up, but opening this draft so we can discuss it.

<details><summary>/boot/status.ini corruption in release-valdiation tests without 2414fd33e</summary> 
image(1)
image(2)
</details>

### DIFF
--- a/base/self-update/default.nix
+++ b/base/self-update/default.nix
@@ -1,8 +1,28 @@
 { config, pkgs, lib, ... }:
 let
   cfg = config.playos.selfUpdate;
+
+  # propagate status.ini update to legacy systems
+  postInstallHandler = pkgs.writeShellApplication {
+    name = "post-install";
+    runtimeInputs = with pkgs; [ coreutils ];
+    text = ''
+        tmpfile=$(mktemp)
+        trap 'rm -f $tmpfile' EXIT
+
+        cp -av /var/lib/rauc/status.ini "$tmpfile"
+        # Ensure /boot/status.ini is always order than /var/lib/rauc/status.ini
+        # to avoid statusfile-recovery.service from overwriting it. See below
+        # for details.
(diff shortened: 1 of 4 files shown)


yfyf:
Looking at this again with #329 about to land, I think this approach could work, especially since we pin down RAUC and deps. The benefit is that it safeguards us against FAT corruption during installs, which are bound to happen, particularly as we plan to have more standalone / at-home installations, where PlayOS is more likely to be powered on only during use.

The key idea is:
- Storing RAUC status file in the persistent data volume (`/var/lib/rauc`), which uses a proper journaling FS (ext4)
- Using `/boot/status.ini` only as a last-resort recovery storage (e.g. if persistent data is wiped or corrupted)

The tricky part is ensuring backwards compatibility with older versions. This PR handles that directly by "recovering" from `/boot/status.ini` if it has been updated by the other slot. This complicates the recovery logic. An alternative approach could be to do this change in stages:
- Release a PlayOS version that only mirrors the RAUC status file to `/var/lib/rauc/`, but still uses `/boot/` as primary source
- Once all installations have progressed to write to `/var/lib/rauc`, we can switch to using it as the primary source and only recover if it is missing. This would simplify the recovery logic by a lot.

PR #297, opened 2025-12-15
Activate hardware watchdog

knuton:
This lets systemd activate the hardware's watchdog device, meaning systemd will control and periodically touch the watchdog. If systemd or even the Linux kernel should end up crashed or frozen, the watchdog would perform a reset at the hardware level, such that the system reboots. systemd pings the watchdog at half the set interval.

We use the default watchdog device, `/dev/watchdog0`.

This is the minimum configuration to utilize the watchdog with systemd. Beyond this, if there is any concern that individual services may fail to function while the system is generally operational (but maybe in a bad state), systemd also allows integrating individual services with the watchdog, such that both repeated failure of the service and failure to call the watchdog will trigger a system reboot.

I verified that the configuration does what it says by triggering a kernel panic in the QEMU VM (`echo c > /proc/sysrq-trigger`), which reboots with watchdog active and just stays panicked otherwise. The same does not work on a vanilla install of PlayOS, as already configured to reboot on panic. An alternative test might be loading a custom kernel module that hangs.

Pondering automated tests -- it would currently just sanity check that the feature is really enabled. Also pondering whether to move to base instead of application.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [x] User manual updated

### DIFF
--- a/application.nix
+++ b/application.nix
@@ -94,10 +94,14 @@ rec {
 
       # Limit virtual terminals that can be switched to
       # Virtual terminal 7 is the kiosk, 8 is the status screen
-      playos.xserver.activeVirtualTerminals = [ 7 8 ];
+      playos.xserver.activeVirtualTerminals = [ 1 7 8 ];
+      users.users.root = {
+        hashedPasswordFile = null;
+        initialHashedPassword = "";
+      };
 
       # System-wide packages
-      environment.systemPackages = with pkgs; [ breeze-contrast-cursor-theme ];
+      environment.systemPackages = with pkgs; [ breeze-contrast-cursor-theme gdb ];
 
       # Avoid bloating system image size
       services.speechd.enable = false;
@@ -284,7 +288,7 @@ rec {
       # greeting uninformed users. 60 s is short enough so the kiosk may never freeze for long,
       # but long enough to rule out false positives due to any brief moments of excessive load.
       systemd.watchdog = {
-        runtimeTime = "60s";
+        runtimeTime = "30s";
       };
       systemd.services."watchdog-loader" =
       let


yfyf (approved):
Looks good, I like the idea of avoiding a frozen system. I wonder if there are pitfalls, but I think this realistically can only be determined in the field.

> Pondering automated tests -- it would currently just sanity check that the feature is really enabled.

Not sure we can do much automated testing here. Any setup I tried to think of just ends up testing the implementation of the kernel or systemd, rather than something about our setup. The devil is also in the hardware<->kernel interaction.

> I verified that the configuration does what it says by triggering a kernel panic in the QEMU VM (echo c > /proc/sysrq-trigger), which reboots with watchdog active and just stays panicked otherwise. The same does not work on a vanilla install of PlayOS, as already configured to reboot on panic.

Maybe [`gdp -p 1` to pause systemd](https://unix.stackexchange.com/a/694473) works?

Alternatively, disable the reboot-on-panic and try it without it.

> Also pondering whether to move to base instead of application.

Probably not? The timeout value is kind of application specific I suppose.

knuton:
> Not sure we can do much automated testing here. Any setup I tried to think of just ends up testing the implementation of the kernel or systemd, rather than something about our setup. The devil is also in the hardware<->kernel interaction.

Agree. At the current complexity of setup (us just enabling the systemd feature), it would be almost misleading to have automated tests, because the big question is not whether the feature works in a configured VM, but whether it practically works with the hardware watchdog on the supported machines.

> Maybe [`gdp -p 1` to pause systemd](https://unix.stackexchange.com/a/694473) works?

Thanks, hadn't come across that, will try.

> > Also pondering whether to move to base instead of application.
> 
> Probably not? The timeout value is kind of application specific I suppose.

:+1:

knuton:
For testing, apply modification below, build `./build shed-key` and perform two trials:

Crash kernel:

1. Boot from USB
2. Go to virtual terminal 1, log in as root
3. `echo c > /proc/sysrq-trigger`
4. Wait for > 1 minute and note whether it rebooted

Pause systemd:

1. Boot from USB
2. Go to virtual terminal 1, log in as root
3. `gdb -p 1`
4. Wait for > 1 minute and note whether it rebooted

```diff
diff --git c/application.nix w/application.nix
index b3a4170e..000c47ed 100644
--- c/application.nix
+++ w/application.nix
@@ -86,10 +86,14 @@ rec {
 
       # Limit virtual terminals that can be switched to
       # Virtual terminal 7 is the kiosk, 8 is the status screen
-      playos.xserver.activeVirtualTerminals = [ 7 8 ];
+      playos.xserver.activeVirtualTerminals = [ 1 7 8 ];
+      users.users.root = {
+        hashedPasswordFile = null;
+        initialHashedPassword = "";
+      };
 
       # System-wide packages
-      environment.systemPackages = with pkgs; [ breeze-contrast-cursor-theme ];
+      environment.systemPackages = with pkgs; [ breeze-contrast-cursor-theme gdb ];
 
       # Avoid bloating system image size
       services.speechd.enable = false;
@@ -276,7 +280,7 @@ rec {
       # greeting uninformed users. 60 s is short enough so the kiosk may never freeze for long,
       # but long enough to rule out false positives due to any brief moments of excessive load.
       systemd.watchdog = {
-        runtimeTime = "60s";
+        runtimeTime = "30s";
       };
 
       playos.hardening.enable = true;

```

PR #303, opened 2026-01-05
Upgrade to Nixpkgs 25.11

knuton:
Updated definitions to resolve deprecation warnings, eval errors and build errors.

## Issues

- [x] kiosk segfaults when started locally
- [x] kiosk crashes in `run-in-vm` because QML imports are not found
- [x] tests using `nixos-test-script-helpers` don't run due to mypy errors
- [x] Integration tests pass, currently failing:
  - [x] controller-interface-labeling.nix
  - [x] controller-proxy.nix
  - [x] controller-system-buttons.nix
  - [x] controller-watchdog-controls.nix
  - [x] kiosk-recovery.nix
  - [x] controller-wifi.nix
    - flaky, failing run: https://github.com/dividat/playos/actions/runs/25785981276/job/75739414045
    - historically this test is quite flaky and this could be just another instance which might still be present on `main`
    - possible that some deep wpa_supplicant and/or service ordering changes are triggering a worse path, which cause some backoffs/timeouts. Attempt to harden it further: 3fb4bc0
  - [x] network-watchdog.nix
    - seems to be flaky, need to review
    - details and fix attempt: cd968d8
    - fix insufficient, still flaky: https://github.com/dividat/playos/actions/runs/25717413931/job/75510417749?pr=303
    - seems to no longer flake after cac19ae 
- [x] E2E tests pass
- [x] Release validation tests pass: https://github.com/dividat/playos/actions/runs/25483115066
- [x] Add a message to all TTYs explaining how to return to the graphical session (to ensure the Ctrl+Alt+F7 change does not confuse operators): 1d81d83 

## Checklist

- [ ] Changelog updated
- [x] Code documented
- [x] User manual updated
- [x] Live ISO boot
- [x] Manual install
- [x] Clean up git commits
- [x] Bloat check: compare bundle sizes (`du -BM -sL $(nix-build -A components.unsignedRaucBundle)`)
  - RAUC bundle increased by 400M (+30%) compare to main (from 1323M to 1724M), need to investigate
  - Main 3 bloat contributors:
    1. Skeleton RAUC + deps: adds around ~70MB compressed (~240 MB uncompressed)
    2. linux-firmware increase: adds around ~150MB compressed (~500 MB uncompressed)
    3. gst-plugins-bad (new dep via pyqt6->qtmultimedia->gst-plugins-bad): adds around 50MB compressed, useless plugins (MIDI, TTS..), no way to disable without causing massive Qt rebuild...
  - The rest seem to be package increases.
  - The biggest potential savings are from trimming the linux-firmware image, made a separate issue to track this: https://github.com/dividat/playos/issues/343
- [x] Bloat check: compare installer sizes (`du -BM -sL $(nix-build -A components.installer.isoImage)`)
  - on main: `2252M`
  - here: `3345M`
  - ~49% (~1100MB) increase
  - attributable to the overall +970MB system image increase (see above) + rescueImage is now using different toplevel init due to skeleton split
  - unfortunate situation, but not much we can do here, #343 is still the best remedy
- [x] Read through NixOS 25.05 and 25.11 release notes: https://nixos.org/manual/nixos/stable/release-notes#sec-release-25.11
  - read through manually and also asked Claude to review, nothing beyond what is already changed/fixed.

### DIFF
--- a/base/networking/default.nix
+++ b/base/networking/default.nix
@@ -76,6 +76,15 @@ in
   };
   # Issue 1: Make sure connman starts after wpa_supplicant
   systemd.services."connman".after = wpaSupplicantServices;
+
+  # connman provides /etc/resolv.conf (via its built-in DNS proxy).
+  # Use the standard nss-lookup.target synchronization point so services
+  # that depend on name resolution can order themselves with After=.
+  systemd.services."connman".wants = [ "nss-lookup.target" ];
+  systemd.services."connman".before = [ "nss-lookup.target" ];
+
+  # avahi-daemon needs /etc/resolv.conf to be available before it starts
+  systemd.services."avahi-daemon".after = [ "nss-lookup.target" ];
   # Issue 2: Restart wpa_supplicant (and thereby connman) after rfkill unblock of wlan
   #          This addresses the problem of wpa_supplicant with connman not seeing any
   #          networks if wlan was initially soft blocked. (https://web.archive.org/web/20191211094135/https://01.org/jira/browse/CM-670)
--- a/base/networking/default.nix
+++ b/base/networking/default.nix
@@ -48,6 +48,9 @@ in
     '';
   };
 
+  # connman runs a local caching DNS proxy and manages /etc/resolv.conf
+  networking.resolvconf.enable = false;
+
   networking = {
     hostName = hostName;
 
--- a/base/networking/default.nix
+++ b/base/networking/default.nix
@@ -21,6 +21,8 @@ in
     traceroute # for connectivity
   ];
 
+  systemd.targets.network.wantedBy = [ "multi-user.target" ];
+
   # Set up networking with ConnMan
   # We need to work around various issues in the interplay of connman and
   # wpa_supplicant for this to work.
--- a/testing/integration/controller-proxy.nix
+++ b/testing/integration/controller-proxy.nix
@@ -73,6 +73,10 @@ pkgs.testers.runNixOSTest {
       ];
 
       config = {
+        # work-around to nixos broken network targets when connman is used.
+        # without this, network.target is never activated.
+        systemd.targets.network.wantedBy = [ "multi-user.target" ];
+
         networking.firewall.enable = false;
 
         services.connman = {
--- a/testing/integration/controller-proxy.nix
+++ b/testing/integration/controller-proxy.nix
@@ -96,13 +96,14 @@ pkgs.testers.runNixOSTest {
   };
 
   extraPythonPackages = ps: [
+    ps.playos-test-helpers
     ps.colorama
     ps.types-colorama
   ];
 
   testScript = {nodes}:
 ''
-${builtins.readFile ../helpers/nixos-test-script-helpers.py}
+from playos_test_helpers import TestPrecondition, TestCheck, wait_for_logs, configure_proxy
 
 latest_version = "9.9.9-TEST"
 
@@ -143,10 +144,10 @@ configure_proxy(playos, proxy_url)
 
 ### === Test scenario
 
-with TestCase("Controller uses proxy for captive portal"):
+with TestCheck("Controller uses proxy for captive portal"):
    playos.succeed("curl -f http://localhost:3333/internet/status | grep TEST_CAPTIVE_RESPONSE") 
 
-with TestCase("Controller is able to query the version and initiate download"):
+with TestCheck("Controller is able to query the version and initiate download"):
     wait_for_logs(playos,
         f"Downloading.*{latest_version}",
         unit="playos-controller.service",
--- a/testing/integration/controller-proxy.nix
+++ b/testing/integration/controller-proxy.nix
@@ -73,10 +73,6 @@ pkgs.testers.runNixOSTest {
       ];
 
       config = {
-        # work-around to nixos broken network targets when connman is used.
-        # without this, network.target is never activated.
-        systemd.targets.network.wantedBy = [ "multi-user.target" ];
-
         networking.firewall.enable = false;
 
         services.connman = {
--- a/application.nix
+++ b/application.nix
@@ -358,7 +358,9 @@ rec {
       '';
 
       # Set a low default timeout when stopping services, to prevent the Windows 95 shutdown experience
-      systemd.extraConfig = "DefaultTimeoutStopSec=15s";
+      systemd.settings.Manager = {
+        DefaultTimeoutStopSec = "15s";
+      };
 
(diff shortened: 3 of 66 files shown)


knuton:
Closing for now, will revisit in future cycle with GRUB skeleton pinned.

yfyf:
Re-based on current `main`

yfyf:
>  kiosk segfaults when started locally

@knuton does kiosk still segfault from `nix-shell` for you at  3e9a11d? It segfaults in `run-in-vm` and I believe it did segfault for me on first run in `nix-shell`, but now it doesn't anymore, so I can't reproduce locally at least. The `run-in-vm` segfault seems to relate to graphics drivers, e.g. I can get it to work by changing to `-vga std`.

I suggest to test with `kioskUrl` set to something other than `play.dividat.com`, because there seem to be other issues that might be making QtWebEngine crash/hang specific to it.

yfyf:
Rebased on top of main.

yfyf:
> I suggest to test with `kioskUrl` set to something other than `play.dividat.com`, because there seem to be other issues that might be making QtWebEngine crash/hang specific to it.

Fixed an issue with kiosk going into an infinite reload in 509788d. Strangely, this seems to also get rid of the segfault for me at least. Nevertheless, I also disabled Vulkan in 8513fc4, because it gets rid of some related errors.

yfyf:
The automated tests now pass, remaining efforts should focus on manual testing / verification.
I am parking this to focus on other things for now.

In the meantime, @knuton @dividat-jgu could you check if the following work for you on this branch:
- `./build vm`
- `./result/bin/run-in-vm`
- `./result/bin/run-in-vm --opengl`
- `cd kiosk; nix-shell;  ./bin/kiosk-browser https://play.dividat.com http://localhost:3333`
- same as above, but before starting, do `rm -rf ~/.cache/kiosk-browser/ ~/.local/share/kiosk-browser/`

dividat-jgu:
> In the meantime, @knuton @dividat-jgu could you check if the following work for you on this branch:

All good!

Notes: I had to use `nixGL` to run the kiosk, and also to run the vm with `--opengl`. With `--opengl`, the output was a bit blurry, but I already had that before.

knuton:
- [x] `./build vm`
- [x] `./result/bin/run-in-vm`
- [x] `./result/bin/run-in-vm --opengl`
- [x] `cd kiosk; nix-shell;  ./bin/kiosk-browser https://play.dividat.com http://localhost:3333`
- [x] same as above, but before starting, do `rm -rf ~/.cache/kiosk-browser/ ~/.local/share/kiosk-browser/`

yfyf:
@knuton I believe this is now at a stage where we can move into manual testing.

I think it would be a good idea to perform the standard PlayOS pre-release manual tests. Do you maybe want to take over this part, since you have all the peripheral hardware? :blush: 

Alternatively, we can go through code review and _then_ do the manual tests.

knuton:
> I think it would be a good idea to perform the standard PlayOS pre-release manual tests. Do you maybe want to take over this part, since you have all the peripheral hardware? 😊

Sure

> Alternatively, we can go through code review and _then_ do the manual tests.

No, this is the way.

knuton:
The live ISO is somehow missing after `./build develop`. Have not investigated yet, only checked that it does not happen on main.

```
lrwxrwxrwx 1 root root  111 Jan  1  1970 result/playos-live-2026.3.0.iso -> /nix/store/ljqyyzg8dlvpkq8g493f9za042r5w5jw-nixos-playos-2026.3.0-x86_64-linux.iso/iso/playos-live-2026.3.0.iso
```

This target does not exist. Other links in `result/` point to existing targets.

Did this PR touch anything related, or should I look for random nix failure?

knuton:
Tests with installed system:

- [x]  Wifi connection can be established.
- [x]  Senso connection can be established through link-local LAN (check in Play).
- [x] RFID reader connects
- [x]  Audio output via HDMI works.
- [x]  Proxy is used immediately when configured and after reboot.
- [x]  After rebooting, login to Play is automatic (no login required).

yfyf:
> Did this PR touch anything related, or should I look for random nix failure?

Fixed, see 314e5dd  and 30b85f9. Also added a checked version of `ln -s` so that it fails at build time instead of silently linking to non-existent things: ba853d7

knuton:
Thanks

Tests with live system:

- [x]  Wifi connection can be established.
- [x]  Senso connection can be established through link-local LAN (check in Play).
- [x]  Audio output via HDMI works.

knuton:
This seems functionally good, time for code review then?

base/networking/default.nix
```diff
@@ -71,6 +76,15 @@ in
...
   systemd.services."connman".after = wpaSupplicantServices;
+
+  # connman provides /etc/resolv.conf (via its built-in DNS proxy).
+  # Use the standard nss-lookup.target synchronization point so services
+  # that depend on name resolution can order themselves with After=.
+  systemd.services."connman".wants = [ "nss-lookup.target" ];
+  systemd.services."connman".before = [ "nss-lookup.target" ];
+
+  # avahi-daemon needs /etc/resolv.conf to be available before it starts
+  systemd.services."avahi-daemon".after = [ "nss-lookup.target" ];
```
knuton, line 87: Hm, to some extent I'd think this may better be set near where avahi is set up, on the other hand this is at least providing context.

testing/integration/controller-proxy.nix
```diff
@@ -73,6 +73,10 @@ pkgs.testers.runNixOSTest {
       ];
 
       config = {
+        # work-around to nixos broken network targets when connman is used.
+        # without this, network.target is never activated.
+        systemd.targets.network.wantedBy = [ "multi-user.target" ];
```
knuton: Since base is imported for this VM, why do we need to repeat that config here (from `base/networking`)?

knuton (commented):
Great job!

I found one issue with the power button handler, addressed in https://github.com/dividat/playos/pull/345.

base/networking/default.nix
yfyf: Hmm, yeah, I think it's better to have it here, so that "conman compat patches" are localized. `application.nix` merely requests that `avahi` is available, but it doesn't really "set it up" in any way.

testing/integration/controller-proxy.nix
yfyf: Good point, removed: 5a6f9b3a3908fdf6edd

base/networking/default.nix
knuton: Let's leave as is.

PR #329, opened 2026-03-05
Separate installer and sub-components as skeleton from updatable system

yfyf:
Reviewing the general situation, I believe it makes little sense that we mash together the packages and configuration needed for the runtime system and the installer into one big mess.

The installer really only needs the system image (a tar blob) + system metadata (a bunch of strings) as inputs. It does not care about what packages are in the system or how it is configured.

In fact, it should be the opposite, the runtime system should depend on the specific skeleton/layout produced by the installer, which involves:
- bootloader
- partition layout
- RAUC statusfile paths

This PR attempts to break apart the two systems, paving the path for a clear delineation between "skeleton" and the updatable system.

In particular:
- nixpkgs are no longer shared between installer/skeleton and runtime/updatable system
- nixos `configuration` is no longer shared
- RAUC and dependencies are pinned in the installer/skeleton and the runtime/updatable system re-uses it to ensure identical versions across releases that do not "evolve" (note: not strictly necessary, but since we anyway pin RAUC's version in the runtime system, might as well pin it all the way with the deps to the skeleton version? The resulting bundle size increase is only 50MB)

## TODOs:

- [x] Store the installer/skeleton ref/version in `/boot` so that `playos-diagnostics` can read it easily
- [x] Populate `system.nixos.label` and `environment.etc.os-release` using the systemMetadata 
- [x] Decide if `installer/` to be renamed to `skeleton/` officially :heavy_check_mark: 
    -  Proposal: **YES**, because it encompasses more than just installer (bootloader, etc)
- [x] Decide if we want to version the installer from the get go or not
    -  Proposal: **assign a version name** to it, but do not implement versioning to keep this PR smaller
    -  Decision: **NO**, having a second pseudo-version just adds confusion. We can determine the skeleton version from initial installer version == initial PlayOS version, which is pinned to a git ref.
- [x] Decide if installer gets it's own git repo / versioning
     - Proposal: **NO**, as a follow-up from above. Can be done in next iterations.
- [x] Decide if installer becomes a flake
     - Proposal: **NO**, also follows- from above
- [x] Decide on the pinned nixpkgs version in installer/skeleton
     - Proposal: set the nixpkgs version to the current pin in `pkgs/default.nix` (`24.11`), which is used in PlayOS since the 2025.3.1 release. The alternative of setting it to the oldest known would introduce issues with hardware compatibility (see e.g. [boot issues with 2022.4.0](https://www.notion.so/dividat/2026-3-0-3106ed7e605280f0be14fc1566f316c2?v=9159df07650b48178aba639beca16de1&source=copy_link#3106ed7e605280a1829bc3c9c18f9077)) and it is quite tricky to even decide how far back into the past to go ([initial commit with nixpkgs pin is from 2018!](https://github.com/dividat/playos/blob/6bcc3c89c16696e292f209d5f7cb65a25b699a6e/nixpkgs.nix)).
 - [x] (Follow-up from previous) For testing purposes, create a release disk image with an older nixpkgs version and add it to the release-validation matrix. Version candidate for this: `2020.7.0` - oldest release containing a "modern" PlayOS [grub.cfg with `save_env ORDER`](https://github.com/dividat/playos/commit/e3565cc97074eac7f5607), probably very few live installations existing with older versions than this.
    - https://github.com/dividat/playos/pull/334
 - [ ] Install an old skeleton version into one the validation machines, so that potential hardware related incompatibilities are discovered.


knuton (commented):
Very good thinking, this seems like a great approach and conceptual cleanup!

installer/pkgs/default.nix
```diff
@@ -0,0 +1,17 @@
+let
+  nixpkgs = builtins.fetchTarball {
+    # nixos-24.05 2024-10-07
```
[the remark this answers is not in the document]
yfyf: 03868d25c1a633f1fe2

installer/install-playos/install-playos.py
```diff
@@ -221,6 +223,17 @@ def install_bootloader(disk, machine_id):
     shutil.copy2(RESCUE_SYSTEM + '/kernel', '/mnt/boot/rescue/kernel')
     shutil.copy2(RESCUE_SYSTEM + '/initrd', '/mnt/boot/rescue/initrd')
 
+    # Mark the skeleton version and installation date
+    with open("/mnt/boot/playos-skeleton.txt", "w") as f:
+        metadata = {
+            "skeleton_version": SKELETON_VERSION,
```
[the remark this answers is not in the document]
yfyf: Removed in dd431a4a1601

pkgs/default.nix
```diff
@@ -13,7 +13,8 @@ let
 
       importFromNixos = path: import (nixpkgs + "/nixos/" + path);
 
-      rauc = (import ./rauc) super;
+      # use RAUC from installer/skeleton
+      rauc = (import ../installer/pkgs).rauc;
```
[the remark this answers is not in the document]
yfyf: Keeping both RAUC and deps pinned to minimize drift across versions.

skeleton/configuration.nix
```diff
@@ -1,10 +1,28 @@
-{ config, pkgs, lib, install-playos, version, safeProductName, fullProductName, greeting, squashfsCompressionOpts, ... }:
+{ install-playos, squashfsCompressionOpts, systemMetadata, ... }:
 
+
+{ modulesPath, lib, ... }:
 with lib;
+with systemMetadata;
+let
+    # TODO: WIP ^_^
+    greeting = label: ''
```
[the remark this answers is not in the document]
yfyf: Attempt at something: 52653ffb 

Here's what it looks like in the TTY:
image(5)

Alternatively, we can just use something more "official" and straightforward, like `figlet "Dividat"` or `figlet "PlayOS"`.

Note that this only changes the _installer_ greeting, not the main one. I am mostly doing this to avoid passing through the `greeting` function, to avoid an extra useless dependency.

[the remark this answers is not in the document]
dividat-jgu: Cute!

dividat-jgu:
I’m having an issue running `nix-build --arg kioskUrl https://shed.dividat.com/`:

```
       error: undefined variable 'installer'
       at /home/joris/code/playos/default.nix:90:42:
           89|       live = if buildLive then live else "";
           90|       installer = if buildInstaller then installer.isoImage else "";
             |                                          ^
           91|     };
```

yfyf:
> I’m having an issue running `nix-build --arg kioskUrl https://shed.dividat.com/`:
> 
> ```
>        error: undefined variable 'installer'
>        at /home/joris/code/playos/default.nix:90:42:
>            89|       live = if buildLive then live else "";
>            90|       installer = if buildInstaller then installer.isoImage else "";
>              |                                          ^
>            91|     };
> ```

Seems like I missed an instance while renaming installer->skeleton, should be fixed with 9f6b5f59

dividat-jgu:
There is a desync between the symlink and the file being written:

I built with `nix-build --arg kioskUrl https://shed.dividat.com/ --arg buildLive false`.

```sh
$ ls -l result/playos-installer-2026.3.0-VALIDATION.iso
lrwxrwxrwx 1 root root 139 Jan  1  1970 result/playos-installer-2026.3.0-VALIDATION.iso -> /nix/store/3gvfp5308027x0pqahbwwkbqx629zl4a-playos-installer-installer-2026.3.0-VALIDATION.iso/iso/playos-installer-2026.3.0-VALIDATION.iso
$ ls -lh /nix/store/3gvfp5308027x0pqahbwwkbqx629zl4a-playos-installer-installer-2026.3.0-VALIDATION.iso/iso/playos-installer-2026.3.0-VALIDATION.iso
ls: cannot access '/nix/store/3gvfp5308027x0pqahbwwkbqx629zl4a-playos-installer-installer-2026.3.0-VALIDATION.iso/iso/playos-installer-2026.3.0-VALIDATION.iso': No such file or directory
$ ls -lh /nix/store/3gvfp5308027x0pqahbwwkbqx629zl4a-playos-installer-installer-2026.3.0-VALIDATION.iso/iso/playos-installer-installer-2026.3.0-VALIDATION.iso
-r--r--r-- 1 root root 2.2G Jan  1  1970 /nix/store/3gvfp5308027x0pqahbwwkbqx629zl4a-playos-installer-installer-2026.3.0-VALIDATION.iso/iso/playos-installer-installer-2026.3.0-VALIDATION.iso
```

The file being generated has 2 `-installer-installer`, and the symlink has only one `-installer`.

dividat-jgu:
I did a manual check to verify that I can produce an installer, install PlayOS in a VM, and start the system and the kiosk browser. :heavy_check_mark:

yfyf:
> The file being generated has 2 `-installer-installer`, and the symlink has only one `-installer`.

Thanks, good catch. I have fixed the double-suffixing in 89713c9e, but also removed the hard-coded ISO path in 5c1bcd8a1f6 (the various hard-coded `$out/foo` paths has to be one of the most annoying things in nix...).

Btw, not sure if you know, but you can use `QUICK_BUILD=1` to test these much changes faster.

dividat-jgu:
> Btw, not sure if you know, but you can use QUICK_BUILD=1 to test these much changes faster.

Thanks for the tip, I didn’t see that one passing. :ok_hand:

skeleton/default.nix
```diff
@@ -0,0 +1,61 @@
...
+
+
+  # Installation script
+  install-playos = pkgs.callPackage ./install-playos {
+    grubCfg = ./bootloader/grub.cfg;
+    inherit rescueSystem;
+    inherit systemImage systemMetadata;
+  };
+
+  configuration = (import ./configuration.nix) {
```
dividat-jgu: `configuration` and `skeleton/configuration.nix` may now lack the indication that it’s the installer configuration, and not the skeleton configuration, which is unclear? Something like: `installerConfiguration = (import ./installer/configuration.nix) {`.

skeleton/default.nix
```diff
@@ -0,0 +1,61 @@
...
+    inherit install-playos squashfsCompressionOpts;
+    inherit systemMetadata;
+  };
+
+
+  isoImage = (pkgs.nixos {
```
dividat-jgu: Same here, `installerIsoImage`?

skeleton/default.nix
yfyf: True. I guess the correct thing to do then is to move all installer things under `skeleton/installer`,   like so:
- `skeleton/install-playos` -> `skeleton/installer/install-playos` (or even `skeleton/installer/script`?)
- `skeleton/configuration.nix` -> `skeleton/installer/configuration.nix`
- `skeleton.isoImage` -> `skeleton/installer/default.nix` 
?

It creates more files/modules, but maybe indeed makes things clearer.

dividat-jgu: 👍

yfyf: Changed in d2d01113.

I kept `install-playos` called as is, because it is quite spread out over the codebase and I didn't want to introduce extra naming changes, the skeleton change is already a big conceptual change.

skeleton/default.nix
yfyf: Same, changed in d2d01113

skeleton/installer/default.nix
```diff
@@ -0,0 +1,25 @@
+{ pkgs, rescueSystem, systemMetadata, systemImage, squashfsCompressionOpts }:
+let
+  # Installation script
+  install-playos = pkgs.callPackage ./install-playos {
+    grubCfg = ../bootloader/grub.cfg;
```
dividat-jgu: It may be fine, but I’m feeling like we’re abusing a bit by picking in the parent directory. Could also be passed from arguments. What do you think?

yfyf: Don't have a strong opinion about this, changed in: 54320532

dividat-jgu:
Looking good on my side, I checked that I could run `QUICK_BUILD=1 nix-build --arg kioskUrl https://shed.dividat.com/ --arg buildLive false`.

We could wait for the remaining point to merge the PR:

> Install an old skeleton version into one the validation machines, so that potential hardware related incompatibilities are discovered.

Or rather track it somewhere and merge now, as it’s not strictly necessary to wait for that?

dividat-jgu (approved):

yfyf:
Added Changelog entry and re-based on top of main.

> We could wait for the remaining point to merge the PR:
> 
> > Install an old skeleton version into one the validation machines, so that potential hardware related incompatibilities are discovered.
> 
> Or rather track it somewhere and merge now, as it’s not strictly necessary to wait for that?

I don't think it's necessary to wait, since this is not directly related to the change in the PR, more a general thought on how to have more coverage in validation.

yfyf:
I will perform a final manual install+update test on hardware and will merge then.

skeleton/installer/configuration.nix
```diff
@@ -1,10 +1,28 @@
-{ config, pkgs, lib, install-playos, version, safeProductName, fullProductName, greeting, squashfsCompressionOpts, ... }:
+{ install-playos, squashfsCompressionOpts, systemMetadata, ... }:
 
+
+{ modulesPath, lib, ... }:
 with lib;
+with systemMetadata;
+let
+    # TODO: WIP ^_^
+    greeting = label: ''
```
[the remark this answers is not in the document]
yfyf: Unfortunately the installer boots into a minimal system that does not have a font which supports unicode:
image

Will re-work it...

yfyf:
> I will perform a final manual install+update test on hardware and will merge then.

Manual test successful:
- `./build master` with this PR, but version set to `2026.3.0-SKELETON`  
- installed on PlayOS PC
- PlayOS upgraded to 2026.3.0

but discovered an issue with the ASCII art that needs to be fixed.

skeleton/installer/configuration.nix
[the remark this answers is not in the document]
yfyf: There was also a double "Installer installer". Fixed view (QEMU screenshot, realized I don't need to flash the USB :grin:)

image

Pro-tip: `qemu-system-x86_64 -cdrom ./result/iso/playos-installer-2026.3.0.iso -cpu max -smp 4 -m 1024`

yfyf:
All done from my side, @dividat-jgu please merge if you have no more comments.

dividat-jgu:
There was a failing integration test, that worked when re-running it. Merging!

PR #336, opened 2026-04-17
Keep the greeting art under 80 char width to avoid wrapping

yfyf:
Cosmetic fix.

While booting, the terminal seems to have a max width of 80 chars and wraps, which causes the ASCII art to render badly


dividat-jgu (approved):
I checked on the VM that the label was on 2 lines before, and that’s fixed with this change. 👌

PR #338, opened 2026-04-23
Add "portable" PlayOS flavour

yfyf:
Meant to be installed on laptops, used for Flex device demos via distributors.

Main current issue is sound. The fixes applied here seem to make sound on my laptop (tested via live ISO) work, however users do not have any way control the sound volume, since the media keys are not mapped to anything.

**TODOs:**
- [x] Add+test volume media key support
- [x] Maybe reduce update download limit to 1MB/s or so? To minimize interference in the scenario where download starts during a demo. Requires making it configurable in `controller`.

### DIFF
--- a/flavors/portable.nix
+++ b/flavors/portable.nix
@@ -1,0 +1,61 @@
+let
+  defaults = import ../application.nix;
+in
+{
+    inherit (defaults) safeProductName version overlays greeting;
+
+    fullProductName = "${defaults.fullProductName} (portable)";
+
+    module = { config, lib, pkgs, ... }: {
+      imports = [
+        defaults.module
+      ];
+
+      # use a lower download limit to ensure minimal interference
+      playos.controller.downloadLimit = "2M";
+
+      # we do not expect a stable/permanent network connection
+      playos.networking.watchdog.enable = lib.mkForce false;
+
+      # metrics are optimized for specific hardware and stable setup, probably
+      # not very useful for ad-hoc portable setups
+      playos.monitoring.enable = lib.mkForce false;
+
+      # Do not hard-code HDMI as default
+      hardware.pulseaudio = {
+        extraConfig = lib.mkForce ''
+          # Respond to changes in connected outputs
+          load-module module-switch-on-port-available
+          load-module module-switch-on-connect blacklist=""
+
+          # Prevent PulseAudio from remembering previous muted states.
+          # First, we unload the default restore module, then reload it
+          # explicitly telling it NOT to restore mute states.
+          unload-module module-device-restore
+          load-module module-device-restore restore_volume=true restore_muted=false
+        '';
+      };
+
+      systemd.user.services.force-unmute = {
+        description = "Force unmute PulseAudio on login";
+        wantedBy = [ "default.target" ];
+        after = [ "pulseaudio.service" ];
+
+        path = [ pkgs.pulseaudio ];
+
+        script = ''
+          # Brief pause to ensure PulseAudio has fully initialized its sinks
+          sleep 2
+          pactl set-sink-mute @DEFAULT_SINK@ 0
+
+          # Optional: Force volume to a safe default so it is never at 0%
+          pactl set-sink-volume @DEFAULT_SINK@ 70%
+        '';
+
+        serviceConfig = {
+          Type = "oneshot";
+          RemainAfterExit = true;
+        };
+      };
+    };
+}
--- a/flavors/portable.nix
+++ b/flavors/portable.nix
@@ -57,5 +57,22 @@ in
           RemainAfterExit = true;
         };
       };
+
+      # Add bindings for media keys to allow volume control
+      environment.etc."sxhkd/sxhkdrc".text = ''
+        XF86AudioLowerVolume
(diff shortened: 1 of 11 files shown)


yfyf:
@knuton @dividat-jgu could you please build a live ISO of this, using:

    QUICK_BUILD=1 nix-build --arg buildVm false --arg buildInstaller false --arg buildBundle false --arg buildLive true --arg buildDisk false --arg applicationPath ./flavors/portable.nix

and check on some laptop if audio works "out of the box" when you boot and start using Play?

EDIT: added experimental media key bindings, would be good to also check if volume up/down works.

dividat-jgu:
Having a look this afternoon.

dividat-jgu:
Working well!

- :ballot_box_with_check: I have sound from the laptop
- :ballot_box_with_check: If I plug headphones, I get sounds on the headphones
- :ballot_box_with_check: If I unplug headphones, I get sounds back from the laptop
- :ballot_box_with_check: The pre-selected volume looks good

But:

- :negative_squared_cross_mark: If I plugged an HDMI screen which can play sound, the sound keeps being played from the laptop. What’s the current target? Would it be an idea to use HDMI over the laptop if available with the external screen?

Also, related but not concerning audio, external screens are mirrored, which looks sensible. But as an optimization, we could disable the laptop screen when closing the lid. Or disable the laptop screen when plugging an external screen. I’m guessing that this would reduce the computation need to display the screen, and let it available for other things.

yfyf:
> Working well!

Thank you for testing!
 
>  ❎ If I plugged an HDMI screen which can play sound, the sound keeps being played from the laptop. What’s the current target? Would it be an idea to use HDMI over the laptop if available with the external screen?

Yes, in general this would be nice. However, at the moment the intention is to only use this with a laptop for Play/Flex demo purposes, without an external monitor, so it is not an immediate problem. I have no idea how to make pulseaudio auto-switch to an HDMI profile reliably  - it cannot get it to work on my personal Debian setup and it also doesn't work on PlayOS. Online sources have confusing info, e.g. some sources say it should work out of the box with `module-switch-on-connect`, while [this one provides a complex solution with udev rules](https://askubuntu.com/a/854079) :shrug:

I also don't have a monitor with audio output at hand I could test with. 

We can probably find a solution, but I propose to delay this until the portable hardware configuration matures, since this is very hard to test without targeting specific hardware. There are plans to use a lightweight portable monitor in the future.

> Also, related but not concerning audio, external screens are mirrored, which looks sensible. But as an optimization, we could disable the laptop screen when closing the lid. Or disable the laptop screen when plugging an external screen. I’m guessing that this would reduce the computation need to display the screen, and let it available for other things.

Yes, but same as above, this is not an immediate use case, so I think it's best to avoid any optimizations. It's also not clear this is always desired - e.g. presenter might interact/observe from a different point. I doubt a mirrored extra screen adds much extra load, since all the rendering is already done (especially assuming identical resolutions) for the primary screen.

yfyf:
Audio + media controls seem to work on a Lenovo V15 G5, tested by jvhr

yfyf:
@knuton I propose we actually merge this into `main`, so that we don't need to dig through branches if we need to update this. I could also add a CI workflow that builds the portable flavor as a test.

Alternatively, I can create separate PRs for the non-config changes, but keep the `flavors/` existing only in a branch. However, the non-config changes will look a bit strange when there's no actual usage in the source tree.

flavors/portable.nix
```diff
@@ -0,0 +1,76 @@
...
+
+        XF86AudioRaiseVolume
+            ${pkgs.pulseaudio}/bin/pactl set-sink-volume @DEFAULT_SINK@ +5%
+
+        XF86AudioMute
+            ${pkgs.pulseaudio}/bin/pactl set-sink-mute @DEFAULT_SINK@ toggle
+      '';
+
+      services.xserver.displayManager.sessionCommands = ''
+        ${pkgs.sxhkd}/bin/sxhkd -c /etc/sxhkd/sxhkdrc &
```
knuton, line 75: Document purpose?

knuton (changes requested):
This looks good, can you lightly clean up the commit history, as we're targeting `main`?

> I propose we actually merge this into main, so that we don't need to dig through branches if we need to update this. I could also add a CI workflow that builds the portable flavor as a test.

Seems reasonable, I would definitely add CI job then, so it doesn't silently break.

flavors/portable.nix
yfyf: Is this what you had in mind?  b8e3dd1

yfyf:
> This looks good, can you lightly clean up the commit history, as we're targeting `main`?

Done, re-ordered the commits, squashed into self-contained chunks.

> > I propose we actually merge this into main, so that we don't need to dig through branches if we need to update this. I could also add a CI workflow that builds the portable flavor as a test.
> 
> Seems reasonable, I would definitely add CI job then, so it doesn't silently break.

Added in 7a4cda9

flavors/portable.nix
knuton: I think I had missed that it belonged to the block above. Anyway, this seems good now.

knuton (approved):

PR #339, opened 2026-04-24
Pin-down external GHA to SHA refs to protect against compromise

yfyf:
Been meaning to do this, but this was the final trigger: https://socket.dev/blog/bitwarden-cli-compromised

Doing this for e.g. `actions/checkout` might be overkill, but I think it is more hygienic if it is identical for all `uses:`

### DIFF
--- a/.github/actions/prep-build-env/action.yml
+++ b/.github/actions/prep-build-env/action.yml
@@ -81,14 +81,14 @@ runs:
       cp -r /root/.aws ~/.aws
 
   - name: Install post-run action
-    uses: yfyf/run-and-post-run@main
+    uses: yfyf/run-and-post-run@d4c684a8c8f09138bed7a27adbf1a7f8dae845bf # main, 2025-07-23
     with:
       post: |
         echo "Waiting for cache uploads to finish"
         sudo -i tsp -w || true
         sudo -i tsp -l
 
-  - uses: cachix/install-nix-action@v31.8.4
+  - uses: cachix/install-nix-action@0b0e072294b088b73964f1d72dfdac0951439dbd # v31.8.4
     with:
       nix_path: nixpkgs=channel:nixos-unstable
       extra_nix_config: |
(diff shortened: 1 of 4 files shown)


yfyf:
Whoops, short SHA's don't work, need to fix.

knuton (approved):
Confirmed that the original refs resolve to those commits with an ad hoc AI-gen bash script:

```bash
#!/bin/bash

# Check if an argument was provided
if [ -z "$1" ]; then
  echo "Usage: $0 <user/repo@sha>"
  echo "Example: $0 torvalds/linux@1bd9238"
  exit 1
fi

# Check if jq is installed
if ! command -v jq &> /dev/null; then
  echo "Error: 'jq' is required but not installed. Please install it (e.g., apt install jq, brew install jq)."
  exit 1
fi

# Parse the input format (user/repo@sha)
# IFS (Internal Field Separator) splits the string at '/' and '@'
IFS='@/' read -r user repo sha <<< "$1"

if [ -z "$user" ] || [ -z "$repo" ] || [ -z "$sha" ]; then
  echo "Error: Invalid format. Please use user/repo@sha"
  exit 1
fi

echo "Fetching commit info for $user/$repo at $sha..."
echo "--------------------------------------------------"

# Fetch data from GitHub API
response=$(curl -s -H "Accept: application/vnd.github.v3+json" \
  "https://api.github.com/repos/$user/$repo/commits/$sha")

# Check if the API returned a "Not Found" message
message=$(echo "$response" | jq -r '.message // empty')
if [ "$message" = "Not Found" ]; then
  echo "Error: Commit or repository not found."
  exit 1
fi

# Extract and print relevant information using jq
echo "$response" | jq -r '
  "Author:  " + .commit.author.name + " <" + .commit.author.email + ">",
  "Date:    " + .commit.author.date,
  "URL:     " + .html_url,
  "--------------------------------------------------\n" +
  .commit.message
'
```

PR #341, opened 2026-04-27
Add missing network error page and installer checklist items

yfyf:
Somehow we forgot the usual Checklist in #330 and need to update screenshot after #329


yfyf:
Err, one sec, let me update the installer screenshot in the manual too while I'm here.

knuton (approved):
I checked that the screenshots (which have different resolution than the ones they replace) render well in the user manual PDF, they do.

PR #346, opened 2026-08-11
Separate nameservers form

knuton:
Triggered by a support call, this separates configuration forms for DNS and static IP.

We introduced the ability to configure nameservers along with configuring static IP, since static IP normally requires configuring nameservers, too. But the converse is not true, it is very sensible to set custom nameservers in combination with obtaining IP via DHCP.

The current bundling forces setting a static IP just to work around some DNS issue, which makes things more challenging for users and introduces unnecessary coupling/brittleness.

We infrequently have customer installations where the DHCP-configured DNS server is causing problems in some way, especially with CloudFront distributions (e.g. issues in working with IPv6/IPv4; improper handling of [truncated DNS UDP packets](https://blog.apnic.net/2024/07/15/revisiting-dns-and-udp-truncation/)). Being able to just set another DNS server, without affecting DHCP would make resolving these much smoother and easier to instruct for local IT departments.

## Checklist

-   [x] Changelog updated
-   [x] Code documented
-   [ ] User manual updated

### DIFF
--- a/controller/Changelog.md
+++ b/controller/Changelog.md
@@ -8,6 +8,7 @@
 # Changed
 
 - os: Pin versions of permanent components produced by installer
+- controller: Allow configuring static IP and DNS servers separately
 
 # [2026.3.0] - 2026-04-22
 
--- a/controller/server/gui.ml
+++ b/controller/server/gui.ml
@@ -386,23 +386,29 @@ module NetworkGui = struct
               m "disabling static ip %s" (get_prop "static_ip_address")
           )
         in
-        let%lwt () = Connman.Service.set_dhcp_ipv4 service in
-        let%lwt () = Connman.Service.set_nameservers service [] in
-        return ()
+        Connman.Service.set_dhcp_ipv4 service
     | Some _ ->
         let address = get_prop "static_ip_address" in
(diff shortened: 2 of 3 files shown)


knuton:
User manual update after change considered OK.

dividat-jgu (approved):
Looking good: in the VM, I can configure either a manual IP, or custom domain name servers. I can see that the config correspond to the input I endered.

I didn’t test that the values were used afterward. For some reason, entering random numbers still indicate that the connection is working. Did you have a check with a custom key on a machine?

knuton:
> I didn’t test that the values were used afterward. For some reason, entering random numbers still indicate that the connection is working. Did you have a check with a custom key on a machine?

I think you might just have observed the machine's/browser's local DNS cache.

If you either request an URL you haven't before or wait a bit, you should see connections failing if you enter e.g. `127.0.0.2` as DNS address. Then on entering `9.9.9.9` or disabling the custom DNS, you should see connections succeeding again.

You could also socat into the VM and `curl` something novel after setting bad custom DNS and it should not work, and should work with good custom DNS.

dividat-jgu:
Can’t reproduce it in the VM, but as you tested in on a USB stick, I think that’s enough. Merging!

PR #349, opened 2026-08-25
Refactor network page in PlayOS settings

yfyf:
Solves https://app.notion.com/p/dividat/Friendly-display-of-MAC-addresses-in-PlayOS-2f06ed7e6052800da9cbeb588c7913a9?source=copy_link

See individual commit messages for details.

Tested both with a mocked setup and a live ISO (on my own PlayOS and one in Dividat HQ).

Note: will require manual updates with new screenshots, but will do it at the end after the review is finished.

### Mocked data

image

### Real PlayOS PC with Senso connected

IMG_1014

## Checklist

-   [x] Changelog updated
-   [X] Code documented
-   [x] User manual updated


dividat-jgu:
Nice information showing up! :ok_hand:

I have some questions on the interface:

- What is the purpose of the border?

image

- Why is there sometimes a chevron and sometimes not?

image

- Checkmark means connected and cross means disconnected or disabled? Is it clear enough? Should we consider a label?

image

- Add more separation between the 2 tables?

image

yfyf:
>   * What is the purpose of the border?

This is the original styling of labels (annotations), just re-using it for network names (which are labels too)

>   * Why is there sometimes a chevron and sometimes not?

Chevron is only present when you can expand/open i.e. there is an `<a href>` link. Disconnected interfaces (no service) cannot be opened.

>   * Checkmark means connected and cross means disconnected or disabled? Is it clear enough? Should we consider a label?

Yes. I think it's fine, it is anyway only "partial" information (e.g. link-local is "connected", but there will be no internet connectivity).

>   * Add more separation between the 2 tables?

You mean between the two `<section>` elements? OK.

dividat-jgu:
> This is the original styling of labels (annotations), just re-using it for network names (which are labels too)

Can an interface be connected to multiple networks? I still think it doesn’t stand well in the table, but if there can be many elements, that’s a solution.

> Chevron is only present when you can expand/open i.e. there is an <a href> link. Disconnected interfaces (no service) cannot be opened.

OK, it looked like from initial testing that I could also open elements without a chevron. But testing it again, it doesn’t look like it’s possible.

---

Nitpicking about the word `Available` in `Available wireless networks`, is it useful to say that it’s available? Are some WIFI networks _unavailable_?

Alternatives: `Wireless networks`, `All wireless networks`, `Visible wireless networks`.

dividat-jgu:
Thanks for the mock. :+1: 

image

Interfaces are not connected though?

yfyf:
> Can an interface be connected to multiple networks? I still think it doesn’t stand well in the table, but if there can be many elements, that’s a solution.

Under normal circumstances - no and the code only assigns at most one service name. But an interface can be connected to a service AND have a Senso connected via the same interface/network (in which case it will have two labels: `Wired` or `FoobarWifi` plus `Dividat Senso`). This is a rare / outlier case, we can of course display them separated with a comma?

> Nitpicking about the word `Available` in `Available wireless networks`, is it useful to say that it’s available? Are some WIFI networks _unavailable_?

It points to the fact that these are not "currently connected". It was originally called this way, I don't see a good reason to change it.

controller/server/view/network_list_page.ml
```diff
@@ -1,53 +1,238 @@
...
+    else
+      annotations @ (service |> Option.map (fun s -> s.name) |> Option.to_list)
+  in
+  let ip_addresses =
+    service
+    |> Option.map service_ip_addresses
+    |> option_filter (fun s -> not (List.is_empty s))
+    |> Option.value ~default:[ "N/A" ]
+    |> List.map txt
+    |> intersperse (br ())
```
dividat-jgu: Using a list of `<div>`, each element should take the all width, and we wouldn’t need to intersperse `<br>`.

yfyf:
> Interfaces are not connected though?

You need to modify the mocked `Ethernet` sections to match your interface names (as returned from `ip link`),
so e.g. change

```
                    'Ethernet': mock_ethernet_dict('wlp0s20f3', 'AA:BB:CC:DD:EE:02')
```

to

```
                    'Ethernet': mock_ethernet_dict('wlan0', 'AA:BB:CC:DD:EE:02')
```

I was thinking that the mock could read the names from `ip link` automatically even, but I was actually mocking the `ip` output in my testing too, so didn't do it.

Edit: Oh, and `State` must be `ready`!

controller/server/view/network_list_page.ml
yfyf: Nice tip, thank you! Done in 5a21779

yfyf:
Updated the docs and default setup in the mock: 35ed12eb

dividat-jgu:
I looked at the network interface from _main_, I don’t see the _label_, maybe it shows in certain circumstances.

image

My suggestions on this:

- Use the same style as _Interface_, making it still more visible than _MAC_ or _IP address_,
- If there are multiple values, either have one per line or use commas (maybe better to have one per line, in case the network contains a comma in its name?),
- Remove the left padding/margin when there is only one element.

image

Should we use just black for addresses? It’s already smaller, maybe enough for reduced visibility?

image

Compared to on _main_, I like that the checkbox has a proper visibility and place.

I’m not completely convinced by how the Network page feels, but it’s a very good improvement other what we previously had, and we can refine in the future if necessary. :+1:

yfyf:
I like your suggestions, thanks, will adapt the style in this way!

yfyf:
> I’m not completely convinced by how the Network page feels

Yeah, it's far from perfect, but implementing these changes I realized two things:
- connman's API and approach to structuring connections is kind of awkward, so it's hard to "fit" things into its model nicely
- striking a good balance between "simple and understandable for non-tech users" and "logical and complete for tech users" is quite hard for networking, since networking setups in general can be... quite messy

yfyf:
How about semi-colons + italic? 

image

The semi-colon should be fine since:
- 99.9% users will never see it if they have a normal PlayOS+Senso setup
- I expect most wifi routers will not allow you to set a name with a semi-colon in it

I feel the italic helps to emphasize that this is label is a _reference_, not just a generic piece of data (like the other fields).

dividat-jgu:
OK for using a semicolon. I would not use italic though, this makes the table to have a lot of different styles. Too much to my taste at least.

yfyf:
@dividat-jgu pushed the styling changes. If it looks good to you now, I will make the screenshots and update the manual.

dividat-jgu (approved):
Looking good, nice job!

yfyf:
@dividat-jgu made some minor styling adjustments and updated the manual - can you have a final look?

dividat-jgu (approved):
All good, I built the manual locally and looked at the new content under the Network section.

